PR #14, opened 2026-04-29
fix: treat rank-tracking max position 0 as "show only unranked"

granata005:
## Summary
- In Rank Tracking → Filters, setting **Desktop position max** or **Mobile position max** to `0` previously returned an empty list (because `position > 0` excluded everything and null positions were dropped first).
- Treat an explicit max of `0` as a sentinel meaning "show only rows where the corresponding position is null (unranked)". Min is ignored in that case.
- Behavior for any non-zero range is unchanged.

## Files
- `src/client/features/rank-tracking/RankTrackingFilters.tsx` — `applyFilters` only.

## Test plan
- [ ] Open a domain with a mix of ranked and unranked keywords.
- [ ] Set Desktop position max = `0` → only rows with `—` for desktop position remain.
- [ ] Set Mobile position max = `0` → only rows with `—` for mobile position remain.
- [ ] Set both to `0` → only rows where both desktop and mobile are `—` remain.
- [ ] Set max = `10` (or any non-zero) → range filter behaves as before.

### DIFF
--- a/src/client/features/rank-tracking/RankTrackingFilters.tsx
+++ b/src/client/features/rank-tracking/RankTrackingFilters.tsx
@@ -163,22 +163,31 @@ export function applyFilters(
     if (excludeTerms.some((t) => kw.includes(t))) return false;
 
     if (filters.minDesktopPos || filters.maxDesktopPos) {
-      const min = filters.minDesktopPos ? Number(filters.minDesktopPos) : 0;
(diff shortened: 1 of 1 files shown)


bensenescu:
Thanks for putting up this PR! I ended up merging this PR which was inspired by this: https://github.com/every-app/open-seo/pull/18

PR #15, opened 2026-04-29
feat: rescan selected rank-tracking keywords with custom depth

granata005:
## Summary
Adds a **Rescan selected** bulk action next to **Remove** in the rank-tracking table. Clicking it opens a small dialog where the user picks a one-off SERP depth (1–10 pages, same control as the config modal) and confirms; the selected keywords are then rescanned at that depth.

The saved `serpDepth` on the config is **not** changed — the override only applies to that single workflow run, so scheduled checks continue with the original depth.

Useful when you want to verify ranks deeper for a specific subset (e.g. unranked keywords) without paying the higher cost on every scheduled check.

## Implementation
- `triggerCheckSchema` accepts an optional `serpDepth` (10..100, multipleOf 10 — same constraints as `createConfig`/`updateConfig`).
- `triggerRankCheck` passes `serpDepth` to `RankTrackingService.triggerCheck`.
- `RankTrackingService.triggerCheck` builds an `effectiveConfig = { ...config, serpDepth }` only for `beginRankCheckRun` when an override is provided. Nothing is written back to the DB.
- Client hook `useRankCheckTrigger.startCheck` accepts `serpDepth?` and forwards it.
- `RankTrackingTable` renders the new bulk action button + a depth-selection modal (with cost estimate via `estimateRankCheckCredits`) and calls a new `onRescanSelected(keywordIds, serpDepth)` prop.
- `RankTrackingDomainDetail` wires the new prop to `startCheck({ keywordIds, serpDepth })`. The button is disabled while another check is running and on the free plan.

Existing billing checks remain authoritative: the paid-plan gate in `triggerRankCheck`, and `serpDepth` is bounded by the same min/max as the config schema.

## Files
- `src/types/schemas/rank-tracking.ts`
- `src/serverFunctions/rank-tracking.ts`
- `src/server/features/rank-tracking/services/RankTrackingService.ts`
- `src/client/features/rank-tracking/useRankCheckTrigger.ts`
- `src/client/features/rank-tracking/RankTrackingDomainDetail.tsx`

## Test plan
- [ ] On a config with `serpDepth = 40` (4 pages), select a few keywords. Click **Rescan selected**, choose `1 page`, confirm.
- [ ] Verify a new run starts (subset run, only the selected keywords). Cost shown in the dialog matches `estimateRankCheckCredits(selectedCount, devices, depth)`.
- [ ] After the run completes, the saved config still shows `serpDepth = 40` (Configure modal unchanged).
- [ ] While a check is running, **Rescan selected** is disabled with the right tooltip.
- [ ] On the free plan, the button is disabled and the existing paid-plan gate on the server fires if bypassed.
- [ ] Bounds: schema rejects `serpDepth` < 10, > 100, or not a multiple of 10.

### DIFF
--- a/src/client/features/rank-tracking/RankTrackingDomainDetail.tsx
+++ b/src/client/features/rank-tracking/RankTrackingDomainDetail.tsx
@@ -404,6 +404,12 @@ function RankTrackingDomainDetailInner({
             domain={config.domain}
             configId={config.id}
             projectId={projectId}
+            devices={config.devices}
+            defaultSerpDepth={config.serpDepth}
+            onRescanSelected={(keywordIds, serpDepth) =>
+              startCheck({ keywordIds, serpDepth })
+            }
+            rescanDisabled={isBusy || isFreePlan}
           />
         </div>
       </div>
--- a/src/client/features/rank-tracking/RankTrackingTable.tsx
+++ b/src/client/features/rank-tracking/RankTrackingTable.tsx
@@ -1,6 +1,6 @@
 import { useState } from "react";
 import { toast } from "sonner";
-import { Loader2, Trash2 } from "lucide-react";
+import { Loader2, RefreshCw, Trash2 } from "lucide-react";
 import {
   flexRender,
   getCoreRowModel,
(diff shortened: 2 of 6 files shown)


bensenescu:
Hey @granata005, I don't think I want to merge this feature because I think this UI / behavior could get overly complicated.

What if Instead of selecting keywords to rescan, you could just override the depth checked for these keywords so that they're picked up in future scans too? 

I think the UI would be similar to what you've done in this PR, but the new depth would be persistent and apply to all future rank check runs.

PR #17, opened 2026-04-29
feat: shift-click range selection for keyword tables

granata005:
## Summary
Click a checkbox or row to set an anchor, then **Shift+click** another checkbox/row to apply the anchor's selection state to every row between them in the table's current sort/filter order. Same gesture used in Gmail, Finder, and most file managers — saves a lot of clicking when you want to remove or operate on a contiguous block of keywords.

The behaviour is symmetric:
- If the anchor is selected, Shift+click extends the selection to the target.
- If the anchor was just deselected, Shift+click clears the selection across the range.

## Implementation
- New helper `tableSelection.ts → applyShiftRangeSelection(e, row, table, anchorRef)` for TanStack Table rows. Returns `true` if it applied a range action (caller skips its default toggle); otherwise just updates the anchor and returns `false`.
- Anchor is a `useRef<string | null>` lifted to each table-host component.
- Wired into:
  - **Rank tracking results table** — checkbox column in `RankTrackingColumns.tsx` (`makeSelectColumn(anchorRef)`), anchor created in `RankTrackingTable.tsx`.
  - **Keyword suggestion step** — checkbox column and full-row click in `KeywordSuggestionStep.tsx`.

Range is computed against `table.getRowModel().rows`, so it respects the user's current sort and any filters — Shift+click between two visible rows always selects what they see between them, regardless of the underlying data order.

## Files
- `src/client/features/rank-tracking/tableSelection.ts` (new)
- `src/client/features/rank-tracking/RankTrackingColumns.tsx`
- `src/client/features/rank-tracking/KeywordSuggestionStep.tsx`

## Test plan
- [ ] In the rank-tracking results table: click first keyword's checkbox, scroll, Shift+click another keyword → all in between get selected.
- [ ] Re-sort the table (e.g. by Position desc), repeat Shift+click → range follows the visible order.
- [ ] Apply a filter, then Shift+click → only visible/filtered rows in the range get toggled.
- [ ] Click an already-selected checkbox to deselect (anchor now unselected), Shift+click another → range gets deselected.
- [ ] In the **Add Domain → Keyword suggestions** step: same gesture works on the checkbox column and on Shift+click anywhere on the row.
- [ ] Plain click (no Shift) toggles a single row exactly as before.

### DIFF
--- a/src/client/features/rank-tracking/KeywordSuggestionStep.tsx
+++ b/src/client/features/rank-tracking/KeywordSuggestionStep.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from "react";
+import { useEffect, useMemo, useRef, useState } from "react";
 import { useMutation, useQuery } from "@tanstack/react-query";
 import {
   useReactTable,
@@ -15,6 +15,7 @@ import { getDomainKeywordSuggestions } from "@/serverFunctions/domain";
 import { addTrackingKeywords } from "@/serverFunctions/rank-tracking";
 import { getStandardErrorMessage } from "@/client/lib/error-messages";
 import { SortableHeader } from "./RankTrackingColumns";
+import { applyShiftRangeSelection } from "./tableSelection";
 
 type SuggestedKeyword = {
   keyword: string;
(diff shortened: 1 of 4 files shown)


bensenescu:
Thanks for putting up this PR!

I ended up merging this PR which was inspired by this: https://github.com/every-app/open-seo/pull/20

I added you as the co author on my commits, but its still not picking you up as a participant for some reason. Maybe a Github bug?

 Happy to try to figure out how to fix that if you want to be an official "contributor". Will try to get to the next PR quicker

PR #46, opened 2026-06-28
feat(rank-tracking): allow explicit language selection

### LINKED ISSUE #34: Allow explicit language selection and editing in Rank Tracking
## Problem
Rank Tracking should let users explicitly choose and edit language. Inferring language from location is not enough because one location can have multiple valid search languages.

## Evidence
- User feedback: "Be able to set language in keyword rank tracking and be able to edit language using ‘Configure’... Location = Canada and Language can be French or English."

## Implementation check
Checked the OpenSEO repo on `main` (`90a04d18`). Relevant files:
- `src/db/app.schema.ts`: rank tracking configs already store `languageCode`.
- `src/types/schemas/rank-tracking.ts` and `src/serverFunctions/rank-tracking.ts`: create/update paths already accept rank tracking config fields.
- `src/client/features/rank-tracking/RankTrackingConfigModal.tsx`: setup/edit currently derives language from country via `getLanguageCode(locationCode)` instead of exposing it.

This is implementation-ready: expose language selection in the existing setup/edit modal and pass the selected `languageCode` through existing config create/update paths.

## Suggested direction
Add a language selector to Rank Tracking setup and Configure/Edit. Keep the current country-derived language as the default, but let users override it.

## Acceptance criteria
- [ ] Users can choose language during Rank Tracking setup.
- [ ] Users can edit language from Configure/Edit for an existing tracker.
- [ ] Existing trackers keep sensible/default language behavior.
- [ ] Users can create different language configurations for the same country when needed.
- [ ] Tests or QA cover country + language combinations such as Canada/English and Canada/French.

Devamol10:
Resolves #34 

### Description
This PR introduces explicit language selection for Rank Tracking configurations, allowing users to override the country's default language while maintaining a seamless default UX.

### Changes Made
- **Language picker:** Added a "Language" dropdown to `RankTrackingConfigModal`, below "Country".
- **Smart Synchronization:** Changing the country resets the language to that country's default.
- **Mutations:** `createRankTrackingConfig` / `updateRankTrackingConfig` submit the explicitly selected `languageCode`.
- **AI Suggestions:** `KeywordSuggestionStep` inherits the user's explicit language choice.

### Testing
- ✅ Verified state synchronization between Country and Language dropdowns.
- ✅ Verified the correct `languageCode` is passed to the API mutations and the AI keyword suggestion step.

---

### Maintainer follow-up (@bensenescu)

Pushed commits on top to fix build blockers, get the language coverage right, and merge `main` in. `pnpm ci:check` is green (prettier, knip, tsc, oxlint).

**What changed**
- **Fixed the modal (was broken / wouldn't compile).** The diff left the old native country `<select>` in place with an empty `onChange={}` (a TSX syntax error) alongside the new `LocationSelect`, so the modal rendered two country pickers. The `LocationSelect` handler also treated its numeric argument as a DOM event (`Number(e.target.value)` → `NaN`). Removed the dead `<select>` and wired the handler correctly.
- **Per-country language list (the main UX fix).** Instead of showing every language for every country, the picker now shows only the languages **DataForSEO supports for the selected country**. Most countries expose just their default; the ~20 genuinely multilingual ones list their real set (US en/es, Canada en/fr, Switzerland de/fr/it, India en/hi, etc.). Built from the DataForSEO Labs `locations_and_languages` endpoint. The Language dropdown is disabled when a country offers a single language.
- **`googleAdsOnly` countries → default only.** The 48 smaller markets have no per-country language data in DataForSEO (Google Ads is one flat global list), so by product decision they show their default language only.
- **Master language list** (`LANGUAGE_OPTIONS`, internal) is sourced live from the DataForSEO SERP `/v3/serp/google/languages` endpoint. Source endpoints are documented in code comments + inline PR comments.
- **Merged `main`** (resolved one trivial conflict in the client locations shim).
- **knip config:** added `scripts/**` to `knip.jsonc` entry — the standalone CLI scripts were already flagged as unused on `main`; bundled the one-line fix so this branch's `ci:check` is green.

**How to review**
1. `RankTrackingConfigModal.tsx` — single `LocationSelect` for Country + one `<select>` for Language; changing country resets language to that country's default; select disables for single-language countries.
2. `getLanguageOptions()` + `MULTI_LANGUAGE_LOCATIONS` in `src/shared/keyword-locations.ts` — the per-country filtering. See the inline comment for the DataForSEO source endpoint.
3. Rank tracking runs against the SERP API, which accepts mismatched-but-valid location+language pairs and fails invalid codes at **zero cost** — so this is a UX restriction, nothing is technically blocked, and there's no charged-but-failed exposure.

**Judgment calls already made (flagging in case you'd revisit)**
- `googleAdsOnly` countries show **default language only** (your call).
- Multilingual country lists come from **Labs** `available_languages` (the only per-country source DataForSEO offers); rank tracking itself is SERP, which would accept more.
- knip fix is bundled here; it's really a `main` hygiene fix riding along.

**Note:** an earlier inline comment on the language list is now marked "outdated" by GitHub because lines shifted — its content (SERP languages source) is still accurate.


src/shared/keyword-locations.ts
```diff
@@ -490,6 +490,150 @@ export const LOCATION_OPTIONS: readonly LocationOption[] = [
...
+ * (Country list above comes from the sibling Labs endpoint cited at the top of
+ * this file: /v3/dataforseo_labs/locations_and_languages.)
+ *
+ * `code` is the DataForSEO `language_code` (authoritative); `label` is its
+ * `language_name`, lightly cleaned for display. Deviations from the raw
+ * endpoint: the deprecated `iw` Hebrew alias and the redundant `no` are
+ * dropped (Norway uses `nb`, which both SERP and Labs accept). Every country
+ * default in LOCATION_OPTIONS must appear here so the picker can show it.
+ */
+export const LANGUAGE_OPTIONS = [
```
bensenescu: Source note for future refreshes: this language list is the full set the DataForSEO SERP (Google) API accepts — refresh it from the live endpoint (auth required): `GET https://api.dataforseo.com/v3/serp/google/languages`. The country list (LOCATION_OPTIONS above) comes from the sibling Labs endpoint: `GET https://api.dataforseo.com/v3/dataforseo_labs/locations_and_languages`. Deviations are intentional: dropped deprecated `iw` (Hebrew alias) and redundant `no` (Norway uses `nb`, valid for both SERP and Labs). Same info is in the code comment just above.

src/shared/keyword-locations.ts
```diff
@@ -515,6 +662,50 @@ export function getLanguageCode(locationCode: number): string {
...
 }
 
+/**
+ * Countries where DataForSEO offers more than one language, from the Labs
+ * locations_and_languages endpoint (each country's default is included).
+ * Every other country offers just its single default (see getLanguageOptions);
+ * googleAdsOnly countries have no per-country language data, so they fall back
+ * to the default too. Keep each list's codes present in LANGUAGE_OPTIONS.
+ */
+const MULTI_LANGUAGE_LOCATIONS: Record<number, readonly string[]> = {
```
bensenescu, line 672: Per-country language data source: this map is built from the DataForSEO **Labs** `locations_and_languages` endpoint (`GET https://api.dataforseo.com/v3/dataforseo_labs/locations_and_languages` → each location's `available_languages[].language_code`). Only countries that expose more than one language are listed; everything else falls back to its single `LOCATION_OPTIONS` default in `getLanguageOptions` below. `googleAdsOnly` countries have no per-country language data in DataForSEO (Google Ads exposes one flat global list), so by product decision they show their default language only.

PR #55, opened 2026-07-02
feat: Search Performance page with striking-distance report (GSC data in the UI)

### LINKED ISSUE #54: Proposal: Search Performance page (GSC data in the UI) with a striking-distance view
## Proposal

Add a **Search Performance** page that renders Google Search Console data inside the project UI, with a **striking distance** view as the headline report. I would like to build this and want to check the direction before opening a large PR.

## Why

The GSC integration is fully implemented server-side (OAuth, `GscService.getPerformance`, the request builder in `searchAnalytics.ts`) but today it is only consumed by the two MCP tools. No route in the app renders a single click or impression. That leaves the roadmap item ("Google Search Console Integration + MCP") feeling incomplete in the UI, and it misses the cheapest value in the product: own-site query data is free, while everything else costs DataForSEO credits.

## Proposed scope

- New route `p/$projectId/search-performance` plus a nav item, visible when GSC is connected; the existing `SearchConsoleConnectionCard` becomes the empty state.
- A server function wrapping the existing `GscService.getPerformance` (28-day default range, previous-period compare).
- Views:
  - Totals: clicks, impressions, CTR, average position, with deltas vs the previous period.
  - Queries and Pages tables with device/country filters (existing table components).
  - **Striking distance**: `dimensions: ["query","page"]`, filtered to average positions 5 to 20, sorted by impressions. This is the "what should I improve next" list, and it costs $0.
- A "Save as keywords" row action that pushes selected queries into the saved-keywords list (existing `saveKeywords` path), so free GSC data feeds the research -> save -> track loop.

## Non-goals (for this PR)

- No changes to rank tracking or its schema. GSC average position is an impression-weighted average across all SERPs and should not be blended into per-location/per-device `rank_snapshots`. If useful later, a display-only GSC column beside the tracker could be a follow-up.
- No new OAuth or settings work; connection management stays where it is.

Happy to adjust scope, naming, or placement to whatever fits your plans for the GSC roadmap item. If the direction looks right I will open the PR within a few days.

mattmacrocket:
Implements the Search Performance proposal from #54 (thanks for the pointers, all three are in).

## What

A **Search Performance** page that puts the project's Google Search Console data in the product for the first time, with a **striking distance** report as the headline view:

- Totals (clicks, impressions, CTR, avg position) with deltas vs the previous same-length period, in the rank-tracking "insights above data" layout.
- **Striking distance**: query x page rows at positions 5 to 20 sorted by impressions, i.e. the queries where a content improvement most plausibly moves traffic. Rows are selectable with a "Save as keywords" bulk action that feeds the saved-keywords list (and from there rank tracking).
- Queries and Pages tables, device/country/date-range filters, CSV and Google Sheets export.
- All first-party GSC data: zero DataForSEO spend. Six `searchAnalytics.query` calls run in parallel per load, comfortably inside GSC quota with the app's 5-minute query cache.
- Disconnected projects render the existing `SearchConsoleConnectionCard` as the empty state; a dead/denied grant (401/403) also routes there via the existing `isExpectedGrantFailure` seam instead of error tracking.
- Standalone nav tab, per your suggestion.

Built on the shared table components (`useAppTable`/`AppDataTable`, `SortableHeader`, `makeSelectionColumn` with shift-range selection, `TableBulkActionBar`, `TableExportMenu`), mirroring the rank-tracking/domain-overview patterns.

## How to review

1. **Server**: `getSearchPerformanceReport` (one server function, six parallel GSC queries) and the pure shaping helpers + unit tests (`searchPerformanceReport.ts`): impression-weighted position, the 5..20 band, previous-period math.
2. **Client**: the feature folder (Page, Parts, Columns). The Columns/Parts split follows the rank-tracking convention.
3. **One shared-file touch**: `SearchConsoleConnectionCard` now also invalidates the `searchPerformance` cache on connect/disconnect, so the page updates immediately after picking a property (also covers connecting from Project Settings).
4. **Judgment calls to sanity-check**:
   - Striking-distance band is fixed at positions 5 to 20 (no UI to tune it in v1).
   - Totals are computed by summing daily rows because `buildSearchAnalyticsRequest` defaults empty dimensions to `["query"]`; a dimensionless aggregate call would mean touching the shared MCP request builder, which I kept out of scope.
   - Filters/range/tab live in `useState` (saved-keywords precedent). If you would rather have them in URL search params like Domain Overview, happy to switch.

## Review notes (known trade-offs left for your judgment)

- The six parallel `getPerformance` calls each resolve the connection and mint a token; a resolve-once batch path in `GscService` would remove that 6x on Google's token endpoint but changes the service surface, so I left it as a possible follow-up.
- React Query's default retry (3) applies to the fan-out on GSC errors; no sibling query overrides retry, so I kept the default. Can add `retry: false` if you prefer being gentler on a 429ing property.
- Country filter labels are raw ISO alpha-3 codes ("USA", "GBR"); the repo has no alpha-3 country-name map today.
- "Save as keywords" uses the schema's default US locale; a saved query surfaced under a non-US country filter would fetch US metrics if later refreshed on the Saved page.
- Tables render up to 250 rows (server-capped) without `TablePagination`.
- A reviewer suggested consolidating the number formatter into a shared util; I checked and the repo has seven feature-local formatters with no shared util or cross-feature imports, so I matched the existing convention instead. A consolidation chore PR could do all seven at once.

## Testing

- Live against a connected `sc-domain:` property: totals + deltas, all three filters (device filter verified to change the data), column sorting, shift-range selection + bulk save (rows verified in the DB), CSV/Sheets export, the connect-card empty state, and the post-connect cache refresh.
- `pnpm ci:check` green (prettier, knip, tsc, oxlint); vitest 416/416 including the new unit tests.

Closes #54.

---

## Follow-up changes (Ben's review pass)

One squashed commit on top of the above, refining the page after live testing. All still first-party GSC data — **no DataForSEO spend on this page.**

**What changed & why**
- **Striking-distance algorithm fix.** It was filtering each query×page row independently, so a query the site already ranks #1 for (via one page) still surfaced its *other* pages sitting at position 6–20 as "opportunities." It now collapses each query to its **best-ranking page** and keeps the query only when that top page is in the 5–20 band — one row per query, no false opportunities.
- **Server-side pagination** for the Queries and Pages tables (GSC `startRow`) so the page scales to large properties, instead of a fixed 250-row cap. Includes a separate full-dataset export path so CSV/Sheets downloads aren't truncated to the visible page.
- **Client-side pagination** for striking distance (still computes 100, shows 50/page) so the table isn't a wall.
- **UX:** the tables now sit on their own card surface with filters + export in one toolbar (previously floating); Queries tab is prefetched so it opens without a spinner; a **Copy keywords** bulk action (newline-joined); removed "View saved"; page-specific empty-state copy.

**How to review**
1. `searchPerformanceReport.ts` — `buildStrikingDistanceRows` is the algorithm change; the unit tests cover the "drop a query whose top page ranks above the band" and "collapse to best page" cases.
2. `serverFunctions/searchPerformance.ts` — the overview call is now slimmed to totals/striking/countries; `getSearchPerformanceTable` (paginated, `pageSize+1` next-page probe) and `exportSearchPerformanceTable` (full dataset) are new.
3. `SearchPerformancePage.tsx` — `tableQueryOptions` single-sources the table query for both the live `useQuery` and the prefetch; client pagination + layout live here.

**Note / judgment call**
- GSC returns no total row count, so pagination shows prev/next + "Page N" (not "of M"). Export is capped at GSC's 1,000-rows-per-call max; going higher would need looped `startRow` fetches — left out of scope.
- An earlier iteration added DataForSEO volume + KD columns; that was **removed** before this push (added metered spend for marginal value) and squashed out of the history, so the diff is clean.

### DIFF
--- a/src/client/components/table/AppDataTable.tsx
+++ b/src/client/components/table/AppDataTable.tsx
@@ -2,6 +2,7 @@ import {
   flexRender,
   getCoreRowModel,
   getExpandedRowModel,
+  getPaginationRowModel,
   getSortedRowModel,
   useReactTable,
   type ColumnDef,
@@ -38,15 +39,20 @@ type UseAppTableOptions<TData> = Omit<
 > & {
   withSorting?: boolean;
   withExpanded?: boolean;
+  withPagination?: boolean;
 };
 
 export function useAppTable<TData>(options: UseAppTableOptions<TData>) {
-  const { withSorting, withExpanded, ...tableOptions } = options;
+  const { withSorting, withExpanded, withPagination, ...tableOptions } =
+    options;
   return useReactTable({
     ...tableOptions,
     getCoreRowModel: getCoreRowModel(),
     ...(withSorting ? { getSortedRowModel: getSortedRowModel() } : {}),
     ...(withExpanded ? { getExpandedRowModel: getExpandedRowModel() } : {}),
+    ...(withPagination
+      ? { getPaginationRowModel: getPaginationRowModel() }
+      : {}),
   });
 }
 
--- a/src/client/features/gsc/SearchConsoleConnectionCard.tsx
+++ b/src/client/features/gsc/SearchConsoleConnectionCard.tsx
@@ -68,6 +68,9 @@ export function SearchConsoleConnectionCard({
       void queryClient.invalidateQueries({
         queryKey: ["searchPerformance", projectId],
       });
+      void queryClient.invalidateQueries({
+        queryKey: ["searchPerformanceTable", projectId],
+      });
     },
     onError: (error) => toast.error(getStandardErrorMessage(error)),
   });
@@ -83,6 +86,9 @@ export function SearchConsoleConnectionCard({
       void queryClient.invalidateQueries({ queryKey: GRANT_STATUS_KEY });
       void queryClient.invalidateQueries({
         queryKey: ["searchPerformance", projectId],
+      });
+      void queryClient.invalidateQueries({
+        queryKey: ["searchPerformanceTable", projectId],
       });
     },
     onError: (error) => toast.error(getStandardErrorMessage(error)),
--- a/src/client/features/gsc/SearchConsoleConnectionCard.tsx
+++ b/src/client/features/gsc/SearchConsoleConnectionCard.tsx
@@ -63,6 +63,11 @@ export function SearchConsoleConnectionCard({
       setPicking(false);
       void queryClient.invalidateQueries({ queryKey: connectionKey });
       void queryClient.invalidateQueries({ queryKey: GRANT_STATUS_KEY });
+      // The Search Performance report caches {connected:false}; refresh it so
+      // the page shows data right after connecting instead of the stale card.
+      void queryClient.invalidateQueries({
+        queryKey: ["searchPerformance", projectId],
+      });
     },
     onError: (error) => toast.error(getStandardErrorMessage(error)),
   });
@@ -76,6 +81,9 @@ export function SearchConsoleConnectionCard({
       // Disconnect can drop the account-level grant server-side; keep the
       // shared grant-status cache (onboarding step + re-engagement nudge) honest.
       void queryClient.invalidateQueries({ queryKey: GRANT_STATUS_KEY });
+      void queryClient.invalidateQueries({
+        queryKey: ["searchPerformance", projectId],
+      });
     },
     onError: (error) => toast.error(getStandardErrorMessage(error)),
   });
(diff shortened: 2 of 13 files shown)


bensenescu:
Hey @mattmacrocket, thanks for the PR. This is a great value add. I made some changes to the UI and the striking distance logic if you want to read the update in the PR description.

Next time it would be helpful to attach a video for these UI heavy changes. 

You should join the discord too!

https://discord.gg/c9uGs3cFXr

mattmacrocket:
Thanks for the merge and the follow-up pass, Ben. The striking-distance change is a real catch: my version filtered each query x page row independently, so a query the site already won via one page could still surface its weaker pages as opportunities. Collapsing to the best-ranking page first is clearly right, and the pagination work reads like the pattern to reuse going forward. Noted on attaching a video for UI-heavy PRs, will do that from now on. See you in the Discord.

PR #59, opened 2026-07-05
feat: add check_site_status MCP tool

subhansh-dev:
Adds a new MCP tool `check_site_status` that performs a lightweight HTTP HEAD/GET request to check if a website is reachable. Returns status code, response time, content type, and content length.

Uses no DataForSEO credits — useful for verifying a domain is live before running paid SEO tools.

**Input:**
- `url` (required): Full URL to check
- `method` (optional, default `head`): HEAD or GET

**Output:**
- `statusCode`, `statusText`, `responseTimeMs`, `contentType`, `contentLength`

### DIFF
--- a/src/server/mcp/server.ts
+++ b/src/server/mcp/server.ts
@@ -23,6 +23,7 @@ import {
   inspectUrlsTool,
 } from "@/server/mcp/tools/search-console-tools";
 import { whoamiTool } from "@/server/mcp/tools/whoami";
+import { checkSiteStatusTool } from "@/server/mcp/tools/check-site-status";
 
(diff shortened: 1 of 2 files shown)


bensenescu:
Hey @subhansh-dev, I'm hesitant to add this for the following reasons:
1. Extra tool call which could confuse agents
2. Potentially flakey (could say site isn't available for lots of reasons, even though it exists)
3. Slow agents down with an extra tool call before each website.

If you're consistently seeing your agent call urls which don't exist, could you send those logs or some examples?

Maybe we can fix this with a skill or by some other means.

subhansh-dev:
fair points. was worth a shot. ill look for something else to contribute

PR #61, opened 2026-07-06
fix(audit): resolve 508 loop-detected errors from trailing-slash redirect cycle

RDeemer63:
## What this fixes

Site audits were returning 508 "Loop Detected" for a large fraction of pages — 29 of 51 pages on a WordPress site during testing. Affected pages showed title "Loading..." and status 508 instead of crawling.

## Root cause

`normalizeUrl()` stripped trailing slashes from discovered links — `/services/` became `/services`. WordPress (and most CMS platforms) canonicalize to the trailing-slash form and 301-redirect the non-trailing version. The crawler queued `/services`, followed the redirect to `/services/`, then `normalizeUrl()` stripped it again, and workerd's redirect-loop detector fired.

The two-pattern breakdown:

**Pattern A — trailing-slash stripping creates the redirect source:**
```
crawl /services/  →  discover link /services/  →  normalizeUrl strips →  queue /services
fetch /services   →  301 to /services/         →  normalizeUrl strips →  queue /services  →  508
```

**Pattern B — post-redirect URL not marked visited:**
```
crawl /about  →  304 to /about/  →  mark /about visited (not /about/)
discover /about/ later  →  not in visited set  →  re-queue  →  508
```

## Fix

- Remove trailing-slash stripping from `normalizeUrl()`. Canonical URLs like `/services/` are now preserved so they queue as-is rather than as their redirect source.
- Add `visited.add(pageResult.url)` after each crawl batch so the post-redirect canonical URL is marked visited (fixes Pattern B).
- Add `canonicalUrlKey()` — a www/http/https-tolerant equality function used for the Lighthouse homepage comparison, which had the same class of redirect-mismatch vulnerability.

## Verification

Tested on a 51-page WordPress site. Before: 29 pages returned 508. After: 50/50 pages returned 200 with real crawl data.

Existing 13 url-utils tests pass. Five new tests added:

```
✓ normalizeUrl — preserves trailing slash on path-only URLs
✓ normalizeUrl — preserves absence of trailing slash
✓ canonicalUrlKey — treats www and non-www as equal
✓ canonicalUrlKey — treats http and https as equal
✓ canonicalUrlKey — preserves trailing slash distinction in path
```

### DIFF
--- a/src/server/lib/audit/lighthouse.ts
+++ b/src/server/lib/audit/lighthouse.ts
@@ -1,4 +1,4 @@
-import { detectUrlTemplate } from "./url-utils";
+import { detectUrlTemplate, canonicalUrlKey } from "./url-utils";
 import type { BillingCustomerContext } from "@/server/billing/subscription";
 import { createDataforseoClient } from "@/server/lib/dataforseo";
 import type { LighthouseResult, LighthouseStrategy } from "./types";
@@ -135,8 +135,14 @@ export function selectLighthouseSample(
   // strategy === "auto": homepage + 1 per URL pattern, capped at 10
   const selected = new Set<string>();
 
-  // Always include the start URL / homepage
-  const startPage = validPages.find((p) => p.url === startUrl);
+  // Always include the start URL / homepage.
+  // Use canonicalUrlKey() on both sides so the comparison survives:
+  //   - trailing-slash redirects  (/services → /services/)
+  //   - www ↔ non-www redirects   (www.example.com → example.com)
+  //   - http → https upgrades
+  const startPage = validPages.find(
+    (p) => canonicalUrlKey(p.url) === canonicalUrlKey(startUrl),
+  );
   if (startPage) selected.add(startPage.url);
 
   // Group by URL template pattern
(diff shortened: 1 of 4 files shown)


RDeemer63:
Good catch, @codex — you're right that the post-redirect `visited.add` in `enqueueDiscoveredLinks` runs after batch selection, so both `/about` and `/about/` can enter the same batch when seeded simultaneously.

I've pushed a two-layer fix in 77835e8:

**1. Pre-dispatch guard (`selectNextCrawlBatch`)** — when a URL is selected for a batch, its trailing-slash counterpart is immediately added to `visited`. This means the moment `/about` is dequeued, `/about/` is blocked for the remainder of the crawl (and vice versa). Whichever form dequeues first gets crawled; `enqueueDiscoveredLinks` then marks the real post-redirect final URL visited as before.

**2. Post-batch dedup** — `crawledBatch` is filtered by `pageResult.url` before pushing to `allPages`, `enqueueDiscoveredLinks`, and `persistCrawlProgress`. This is a safety net for other redirect patterns (e.g., two unrelated URLs that happen to resolve to the same destination), and ensures KV progress stays consistent even if the pre-dispatch guard ever misses a case.

The pre-dispatch guard eliminates the wasted API call; the post-batch dedup is defense-in-depth against duplicate rows regardless of cause.

bensenescu:
@RDeemer63 Thanks for putting up this PR! We overhauled the audit recently and incorporated this fix. Closing since its now a duplicate

Issue #30, opened 2026-06-25
Add Google Analytics support to the common MCP server


bensenescu:
## Problem
Users can access Search Console through the OpenSEO MCP server and want GA4/Google Analytics available in the same agent workflow.

## Evidence
- User feedback: "you have already implemented search console for MCP, do you plan anything for google analytics as well to fetch items at one go from your common mcp server?"

## Suggested direction
Write a short spec for adding GA4 read-only MCP tools, starting with SEO-relevant analytics rather than a generic GA dashboard.

Clarify:
- OAuth scope and setup path, likely `analytics.readonly`.
- Initial tool surface: properties, organic landing pages, page performance, conversions, ecommerce if relevant.
- How GA4 data should combine with Search Console data for opportunity scoring.
- Non-goals for v1, especially GA4 admin/write actions.

## Acceptance criteria
- [ ] Spec defines the initial GA4 MCP tool set and required OAuth scopes.
- [ ] Spec explains how setup differs from Search Console.
- [ ] Spec identifies SEO-specific combined workflows worth building first.
- [ ] Spec calls out test/fixture needs and non-goals.

strelov1:
I'd like to work on this — starting with a written spec (per the spec-first label) before any code.

Plan: draft a `specs/000X-ga4-integration.md` ADR, same shape as `specs/0003-google-search-console-integration.md` (Status/Context/Decision/Rationale/Consequences), covering:

- **Auth**: incremental OAuth grant via a dedicated `google-analytics` Better Auth `genericOAuth` provider (same pattern as `google-search-console`), scope `analytics.readonly`. Open question I'd like your call on: separate grant from the existing GSC connection, or request both scopes on one combined Google connection to avoid a second consent screen? I'd lean toward keeping them separate (mirrors GSC's `allowDifferentEmails` flexibility for agencies connecting a client's account), but want your take before locking it in.
- **Scoping**: one GA4 property mapped per project (`ga4_connections`, unique per project), same shape as `gsc_connections`.
- **Initial MCP tool surface** (read-only, free — GA4 Data API isn't billed):
  - `get_ga4_report` — thin pass-through of the GA4 Data API `runReport`, scoped to SEO-relevant dimensions/metrics (organic landing pages, sessions, engagement rate, conversions) rather than a generic picker.
  - possibly `list_ga4_properties` for the connect flow, mirroring how GSC lists verified sites.
[...]

Issue #28, opened 2026-05-28
[Bug] Infinite redirect loop and 429 Error when BYPASS_EMAIL_VERIFICATION=true


zenfokus:
Describe the bug

When running the application with AUTH_MODE=hosted and BYPASS_EMAIL_VERIFICATION=true, newly registered users fall into an infinite redirect loop between the root route (/) and /verify-email.

This loop happens so rapidly that it exhausts the Better Auth rate limit within seconds, resulting in a 429 Too Many Requests error on /api/auth/get-session.

To Reproduce
Deploy the app (e.g., using the official Docker image ghcr.io/every-app/open-seo:latest).
Set environment variables:

AUTH_MODE=hosted
BYPASS_EMAIL_VERIFICATION=true

Create a new user account with Email/Password.
Attempt to log in.
Observe the browser network tab or server logs: the app continuously bounces between GET / and GET /api/auth/get-session (navigating back and forth between the app and the verify-email page).
Root Cause Analysis
The loop is caused by a mismatch in how session.user.emailVerified is handled across different routes:

useHostedAuthRouteGuard.ts (Client Guard): This guard strictly checks if (!session.user.emailVerified) and forces a navigation to /verify-email. It does not check if BYPASS_EMAIL_VERIFICATION is enabled. Since Better Auth does not automatically set emailVerified = true in the database when verification is bypassed, newly created users have emailVerified = false.

verify-email.tsx (Verification Page): On this page, the code evaluates const bypassEmailVerification = import.meta.env.BYPASS_EMAIL_VERIFICATION === "true";. Since bypass is true, the useEffect skips the waiting state and executes window.location.replace(redirectTo); (which points back to /).

The Loop:

User goes to /.
useHostedAuthRouteGuard sees emailVerified == false -> Navigates to /verify-email.
/verify-email sees bypassEmailVerification == true -> Redirects to / (Full page reload).
Rinse and repeat.
Suggested Fix
There are a few ways to resolve this:

Option A (Frontend Fix): Update useHostedAuthRouteGuard.ts so that it respects the bypass flag. It should not navigate to /verify-email if BYPASS_EMAIL_VERIFICATION is active. (Note: Care must be taken to ensure import.meta.env.BYPASS_EMAIL_VERIFICATION is properly exposed to the client bundle).

[...]

PR #62, opened 2026-07-06
feat(rank-tracking): add city/region-level location targeting

RDeemer63:
## Problem

Rank tracking currently checks keywords at country level only. For local service-area businesses, this systematically understates performance — and can misreport pages as "not ranking" that are actually driving real business.

We self-host OpenSEO for a 26-client local-services SEO agency. Testing against a real client site (land clearing / dirt work contractor in rural Oklahoma) showed the gap:

| Keyword | US national | Local check | GSC 28-day avg |
|---|---|---|---|
| oklahoma pond builders | #28 | #21 (Enid, OK) | 19.4 |
| dozer service near me | not in top 40 | **#2** (Garber, OK) | — |
| pond excavation contractors | not in top 40 | #13 (Enid, OK) | — |
| land clearing services oklahoma | #30 | #38 (Enid, OK) | — |

The "dozer service near me" result is the clearest example: a national rank tracker reports the page doesn't rank at all; a city-level check from the client's home town shows it ranking #2. That's not an accuracy improvement — it's the difference between a page appearing to have zero value versus being the dominant organic result for a high-intent local query.

The fourth row (moving the wrong direction) is intentional and worth knowing: local targeting helps geo-intent queries and hurts broad non-geo-modified category terms. This nuance is embedded in the UI as contextual guidance.

## What this adds

**Location targeting per rank tracking config.** Optional `location_name` field (nullable; `null` = existing country-level behaviour, unchanged). Value is DataForSEO canonical format: `"Enid,Oklahoma,United States"`.

**DataForSEO-backed location autocomplete.** The config modal searches DataForSEO's SERP locations list (`GET /v3/serp/google/locations/{country}`) as the user types. Only locations DataForSEO actually supports appear. Free-text entry was explicitly avoided: a silently wrong location name produces no error until the API call fails — the same class of invisible bug we've been finding throughout this evaluation.

**Context-sensitive UI hint** based on validation findings:
- _Local_ → best for "near me" queries, city/county keywords, service-area pages
- _National_ → best for broad category terms with no location modifier (local targeting can understate these)

**Schema:** nullable `location_name` column on `rank_tracking_configs` (SQLite + Postgres in parity). Migration included. Unique index updated to `(project_id, domain, location_code, location_name)` so one config can target multiple cities for the same domain.

## Implementation notes

- SERP calls route to `location_name` when set, `location_code` otherwise — one conditional spread per call site, no logic duplication
- Autocomplete uses `requireAuthenticatedContext` (not project-scoped middleware) since location search is user-level, not project-level
- All existing configs with `location_name = null` continue to work exactly as before

## Small sample, honest framing

Four keywords, two towns — enough to confirm the mechanism works and surface the geo-intent nuance, not a comprehensive study. The `dozer service near me` result is a concrete, reproducible proof of value. The outlier is as useful as the successes.


RDeemer63:
All three are valid — addressed in `aaf10ce`. Summary per issue:

---

**P1 — Postgres migration**

Added `drizzle-pg/0006_location_name.sql` with the column addition and the updated indexes (see below). Also updated `drizzle-pg/meta/_journal.json` and `drizzle-pg/meta/0006_snapshot.json` to keep the Drizzle migration state consistent. Note: if the project uses `drizzle-kit generate` in CI, the snapshot may need regeneration via `pnpm run db:generate:pg` — the hand-authored one should be correct but Drizzle sometimes diffs whitespace/ordering.

---

**P2 — Allow clearing a local target**

Root cause was `locationName: undefined` being passed from the modal on the update path when switching to National — the service guard `if (input.locationName !== undefined)` skipped the write, leaving the old value in place.

Fix: the update mutation now sends `null` explicitly (`targetingMode === "local" ? locationName : null`). `updateConfigSchema.locationName` is now `.nullable().optional()` to accept it, and the service input type is widened to `string | null`. The existing `!== undefined` guard correctly passes `null` through to the DB write, clearing the column. The create mutation is unchanged — for new records, `undefined` (omit) → DB NULL is correct.

---

**P2 — NULL uniqueness**

Both `drizzle/0029_location_name.sql` and `drizzle-pg/0006_location_name.sql` now use two partial indexes instead of one:

```sql
CREATE UNIQUE INDEX "rank_tracking_configs_national_idx"
  ON "rank_tracking_configs" ("project_id","domain","location_code")
  WHERE "location_name" IS NULL;

CREATE UNIQUE INDEX "rank_tracking_configs_local_idx"
  ON "rank_tracking_configs" ("project_id","domain","location_code","location_name")
  WHERE "location_name" IS NOT NULL;
```

Both `src/db/app.schema.ts` (SQLite) and `src/db/pg/app.schema.ts` (Postgres) are updated to match, using Drizzle's `.where(sql\`...\`)` on the index builder (supported in drizzle-orm ≥ 0.28, project is on 0.45.x).

bensenescu:
@RDeemer63 Thanks for putting this PR up! This is going to make huge difference for local businesses. 

I would love to learn more about how your agency is using OpenSEO. Shoot me an email ben@openseo.so

bensenescu:
Also, I made some tweaks to your PR as a heads up. They mostly fell into: 
- UX improvements like showing local search volume instead of country level search volume in the tables
- Fixing a performance issue since the list of locations is so large

PR #63, opened 2026-07-08
feat(rank-tracking): city/region-level local targeting + city-scoped metrics

bensenescu:
Continues #62 by @RDeemer63 (their commits are preserved on this branch) — city/region-level location targeting for rank tracking, taken from concept to working end-to-end, plus city-scoped keyword metrics so local trackers stop showing misleading national numbers.

## What changed and why

**Local targeting (from #62, fixed up).** Rank tracking previously checked keywords at country level only, which systematically misreports local-service businesses — the motivating example ranks #2 for "dozer service near me" in its home town while a national check says it doesn't rank at all. Configs now take an optional city/region (searchable combobox in the config modal); SERP checks pass the location through on both the live and queued paths. The original PR had three ship-blockers, all fixed here: the locations API rejects country names (the combobox always returned "No locations found"), the D1 migration was missing its drizzle journal/snapshot (the next `db:generate:d1` would emit a failing duplicate), and the hand-written PG journal timestamp predated migration 0005 (drizzle's migrator would have silently skipped it in production).

**Locations data caching.** The DataForSEO locations endpoint returns the full country list (9.5 MB for the US) and has no search parameter — the original implementation fetched and parsed it per keystroke (~3 s each, real heap pressure). The slimmed list (~1.5 MB) is now cached in KV (30-day TTL, hot reads edge-cached), concurrent cold fills are coalesced, and flipping to Local targeting prewarms the cache. (KV was chosen over R2 + a Cache API layer: one primitive, same read profile, and the Cache API is a documented no-op on workers.dev, which self-hosters use.) Warm searches measured at ~50–90 ms server time. `specs/0008-local-rank-tracking-locations.md` records the options considered and the phased picker roadmap.

**City-scoped keyword metrics.** National volume on a city tracker overstates demand by orders of magnitude ("rv storage near me": 135K national vs 70/mo in Pittsburgh). Local configs now fetch volume/CPC/competition from Google Ads scoped to the tracked location, merged with national KD/intent from Labs. The column reads "Local volume" (tooltip and exports name the city), and keywords Google Ads collapses away get explicit nulls rather than leaking stale national numbers. Adds ~$0.09 per metrics refresh.

Verified: `ci:check` green, 577 unit tests (new coverage for the metrics merge and ISO mapping), and the full flow driven in a real browser — combobox search/selection/validation (US + UK), a Pittsburgh config fetching 70/260 local volumes with national KD merged, and city-targeted rank checks returning positions.

## How to review

1. **Schema + migrations** (`drizzle/0029_*`, `drizzle-pg/0006_*`, both `app.schema.ts`): nullable `location_name` with partial unique indexes (national vs local uniqueness). Both migrations are drizzle-kit-generated; journals/snapshots chain-verified.
2. **DataForSEO seam** (`serp-locations.ts`, `serp.ts`, `google-ads.ts`, `keyword-metrics.ts`): the caching layers and the local/national metrics merge are the main judgment-call areas — in particular `mergeLocalAndNationalRows` semantics for keywords Google Ads drops.
3. **Workflow plumbing** (`RankCheckWorkflow`, `rankCheckPaths`, `rankCheckRunGuards`): mechanical `locationName` thread-through; scheduled and manual paths both carry it.
4. **UI** (`SerpLocationCombobox`, `SearchTargetingField`, config modal, columns/exports): the combobox is the most stateful piece; its select/stale/spinner behaviors were browser-verified.

**Deliberately out of scope:** the keyword *suggestion* step still shows national volumes; picker UX improvements (result ranking, cities-first, suggestions) and zip-code search are Phases 2–3 in the plan doc.

## Review notes (unfixed nitpicks + one rejected finding)

From a five-axis review pass (complexity / security / billing / idioms / cruft) — no blockers survived verification; these were judged not worth changing but are listed for a second opinion:

- `location_name` is length-validated but not checked against the fetched geotarget list; a hand-crafted request with a junk string fails at DataForSEO at cost 0, charged to the caller's own project. Server-side validation would ride along with the Phase-3 registry table if built.
- Local metrics fire two metered calls in parallel (`Promise.all`); each soft-checks the same credit snapshot, so a near-empty account can slip one call past the cap. Both spends are captured in Autumn — pre-check only.
- `prewarmSerpLocations` is a thin wrapper around the same fetch as search; kept as an explicit warm hook.
- The `location_name`-vs-`location_code` param selection appears in three call sites; a shared helper felt like over-extraction for a 3-line idiom.
- `SerpLocationCombobox` copy-adapts LocationSelect's keyboard/outside-click handling; a shared combobox primitive is a reasonable future refactor.
- **Rejected review finding** (for transparency): "rewrite the combobox search onto `useQuery` with a debounced key." Rejected because no such idiom exists in the repo (this is the only per-keystroke typeahead; BrandLookupPage queries on submit), and `useQuery`'s caching would break the verified stale-list-cleared-on-select behavior while relocating rather than removing the bookkeeping. Optional follow-up if instant-retype caching is wanted.


bensenescu:
Superseded: all follow-up commits have been pushed directly to #62 (maintainer edit) so the original contribution merges with full attribution. Review there.

PR #89, opened 2026-07-17
fix(search-tabs): select the right-hand neighbor when closing the active tab

bookingseo:
## Bug
Closing the active search tab selects the tab to the LEFT; browser-tab convention — and this repo's own e2e contract (`keyword-research-navigation.spec.ts`: *"closing the active middle keyword tab removes it and selects the next tab"*) — expect the RIGHT-hand neighbor.

In `closeTab`, the array is post-removal, so `tabs[index]` is the right-hand tab that slid into the closed slot — but the code prefers `tabs[index - 1]`. The suite wasn't running anywhere, so the inversion shipped unnoticed.

## Fix
`tabs[index] ?? tabs[index - 1]` — right-hand neighbor first, left as fallback when the last tab is closed.

## Also
- `aria-label="Search tabs"` on the strip's tablist (the app shell has other tablists; this distinguishes them for assistive tech).
- The spec's final tab-count assertion is scoped to that tablist so unrelated app-shell tabs can't inflate the count. Spec passes 3/3 headless with the fix.

Found by wiring this e2e suite into CI on a fork.

### DIFF
--- a/e2e/keyword-research-navigation.spec.ts
+++ b/e2e/keyword-research-navigation.spec.ts
@@ -115,6 +115,10 @@ test.describe("Keyword Research navigation", () => {
     await expect(
       page.locator(`[data-search-tab-id="${closedTabId}"]`),
     ).toHaveCount(0);
-    await expect(page.getByRole("tab")).toHaveCount(2);
+    // Count only search tabs: the app shell has grown other tablists
+    // (Browse/Chat), so a bare role=tab count would include them.
+    await expect(
+      page.getByRole("tablist", { name: "Search tabs" }).getByRole("tab"),
+    ).toHaveCount(2);
   });
 });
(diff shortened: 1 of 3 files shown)


bensenescu:
Thanks for the fix — the neighbor-selection change is correct (verified against the e2e contract: closing the middle tab of three now selects the right-hand tab), and the aria-label + scoped spec assertion are good improvements. All 3 specs pass locally and ci:check is clean.

I pushed one maintainer edit (e3ea65a6): removed the `languageCode` addition to the domain-overview tab query in `SearchTabStrip.tsx`. It wasn't mentioned in the PR description, and it changes behavior — `resolveLabsMarket` on the server deliberately uses the project's chosen language when the request's location matches the project's location; always sending `getLanguageCode(locationCode)` from the client would override that (e.g. a project set to French in Canada would start querying the location default instead). If there's a concrete issue that change was addressing, happy to discuss it in a separate PR.

PR #95, opened 2026-07-20
fix(rank-tracking): treat SQLite timestamps as UTC in client views

### LINKED ISSUE #94: Treat SQLite rank-tracking timestamps as UTC in client views
## Problem

Rank-tracking client views pass database timestamp strings directly to `new Date(...)`. SQLite's `current_timestamp` values are UTC strings shaped as `YYYY-MM-DD HH:mm:ss` with no timezone marker, so browsers interpret them as local time. Postgres values are ISO UTC strings and do not have this problem.

This can shift rank-check points by the browser timezone offset and can render the previous/next calendar day around midnight.

## Evidence on current `main` (`6f2eb05`)

- `src/db/app.schema.ts`: rank-check and snapshot timestamps default to SQLite `current_timestamp`.
- `src/db/pg/app.schema.ts`: documents the SQLite/Postgres timestamp format difference.
- Under `TZ=Europe/Lisbon`, SQLite UTC input `2026-07-20 00:30:00` currently parses as `2026-07-19T23:30:00.000Z`.
- Affected paths include `RankTrackingDetailHeader.tsx`, `RankTrackingDomainList.tsx`, `KeywordTrendModal.tsx`, `RankTrackingOverview.tsx`, and `RankTrackingHistoryMatrix.tsx`.

The dashboard already handles the same invariant in `src/client/features/dashboard/cardParts.tsx::formatDay` by adding `T` and `Z` for SQLite-form timestamps.

## Suggested direction

Add one shared timestamp parser that treats the SQLite shape as UTC and leaves ISO timestamps intact. Use it in rank-tracking chart keys, exports, and date labels.

## Acceptance criteria

- [ ] SQLite `YYYY-MM-DD HH:mm:ss` and Postgres ISO timestamps resolve to the same UTC instant.
- [ ] Rank-tracking chart points and date labels do not shift based on browser timezone.
- [ ] Invalid timestamps degrade safely rather than throwing during render.
- [ ] Tests cover SQLite and ISO forms, including a value near UTC midnight.
- [ ] Storage and scheduling formats remain unchanged.

lanpublications:
## Bug
SQLite's `current_timestamp` default (`YYYY-MM-DD HH:mm:ss`) has no timezone marker, so browsers parse it as local time even though stored values are UTC. Rank-tracking views shift every instant by the browser's UTC offset: under `TZ=Europe/Lisbon`, `new Date("2026-07-20 00:30:00")` resolves to `2026-07-19T23:30:00.000Z`. Displayed dates can land on the wrong calendar day, and chart point keys built from `getTime()` disagree with ISO-formatted values (Postgres returns ISO UTC strings, so the two dialects render differently for the same stored moment).

Fixes #94

## Fix
Extract the normalization the dashboard already uses (`cardParts.tsx::formatDay`) into a shared helper, `src/client/lib/timestamps.ts`:

- `parseTimestampMs` treats bare SQLite timestamps as UTC and passes ISO strings through
- `formatTimestampIso` / `formatTimestampDate` are formatting wrappers that fall back to the raw string for invalid input instead of rendering "Invalid Date"

Swapped in at every string-timestamp parse site in rank-tracking: `RankTrackingOverview` (chart keys), `KeywordTrendModal` (chart keys, history table, CSV export), `RankTrackingHistoryMatrix`, `RankTrackingDetailHeader`, `RankTrackingDomainList`. `formatDay` now delegates to the shared helper.

`RankTrackingTrendChart.formatDateTick` receives an already-numeric tick, so it needed no change. Storage and scheduling are untouched, matching the issue's scope.

## Tests
11 unit tests in `src/client/lib/timestamps.test.ts`: SQLite-as-UTC parsing, ISO passthrough, same-instant equality near UTC midnight, explicit offsets, and invalid-input fallbacks. `pnpm vitest run` passes 714/714 and `pnpm ci:check` is clean.

### DIFF
--- a/src/client/features/dashboard/cardParts.tsx
+++ b/src/client/features/dashboard/cardParts.tsx
@@ -2,6 +2,8 @@
 // the GSC IntegrationCard (rounded-xl, shadow-sm, header row + divider) so
 // the embedded SearchConsoleConnectionCard doesn't read as a different
 // design system.
+import { formatTimestampDate } from "@/client/lib/timestamps";
+
 export function CardShell({
   title,
   stamp,
@@ -96,15 +98,7 @@ export function newLost(value: number | null): string {
 }
 
 export function formatDay(timestamp: string): string {
-  const ms = Date.parse(
-    // SQLite's current_timestamp default has no timezone marker; treat it as
-    // UTC rather than letting the browser parse it as local time.
-    /^\d{4}-\d{2}-\d{2} /.test(timestamp)
-      ? `${timestamp.replace(" ", "T")}Z`
-      : timestamp,
-  );
-  if (Number.isNaN(ms)) return timestamp;
-  return new Date(ms).toLocaleDateString(undefined, {
+  return formatTimestampDate(timestamp, undefined, {
     month: "short",
     day: "numeric",
   });
--- a/src/client/features/rank-tracking/KeywordTrendModal.tsx
+++ b/src/client/features/rank-tracking/KeywordTrendModal.tsx
@@ -109,7 +109,10 @@ export function KeywordTrendModal({
     const keys = new Set<string>();
     for (const p of points) {
       if (p.position === null) {
-        keys.add(`${parseTimestampMs(p.checkedAt)}:${p.device}`);
+        const ts = parseTimestampMs(p.checkedAt);
+        if (!Number.isNaN(ts)) {
+          keys.add(`${ts}:${p.device}`);
+        }
       }
     }
     return keys;
@@ -364,6 +367,9 @@ function buildChartData(
   const byTime = new Map<number, ChartRow>();
   for (const p of points) {
     const ts = parseTimestampMs(p.checkedAt);
+    if (Number.isNaN(ts)) {
+      continue;
+    }
     const row = byTime.get(ts) ?? { checkedAt: ts };
     row[p.device] = p.position === null ? serpDepth : p.position;
     byTime.set(ts, row);
--- a/src/client/features/rank-tracking/KeywordTrendModal.tsx
+++ b/src/client/features/rank-tracking/KeywordTrendModal.tsx
@@ -5,6 +5,11 @@ import { useQuery } from "@tanstack/react-query";
 import { Modal } from "@/client/components/Modal";
 import { buildCsv, downloadCsv } from "@/client/lib/csv";
 import { captureClientEvent } from "@/client/lib/posthog";
+import {
+  formatTimestampDate,
+  formatTimestampIso,
+  parseTimestampMs,
+} from "@/client/lib/timestamps";
 import { getRankKeywordHistory } from "@/serverFunctions/rank-tracking";
 import type { RankKeywordHistoryPoint } from "@/serverFunctions/rank-tracking";
 import { LOCATIONS } from "@/client/features/keywords/locations";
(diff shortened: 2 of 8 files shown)


isatimur (commented):
I reproduced one remaining timezone bug and one invalid-input gap on this branch:

1. **Date labels still vary by browser timezone.** The parser now produces the correct UTC instant, but `formatTimestampDate()` calls `toLocaleDateString(locales, options)` without a default `timeZone`. Under `TZ=America/Los_Angeles`, `formatTimestampDate("2026-07-20 00:30:00", "en-US", { month: "short", day: "numeric" })` returns `Jul 19`; under `TZ=Europe/Lisbon`, `2026-07-20 23:30:00` returns `Jul 21`. That still violates #94's requirement that rank-tracking date labels not shift with browser timezone. Please default to UTC (for example `{ timeZone: "UTC", ...options }`, allowing an explicit caller override) and add an adversarial-zone regression test that exercises the default call-site behavior.

2. **Chart paths still accept `NaN`.** `parseTimestampMs()` intentionally returns `NaN` for invalid input, but `RankTrackingOverview`, `buildChartData`, and the bottom-band key builder feed that value into Recharts / maps instead of omitting the invalid point. The formatting wrappers degrade safely, but the chart call sites do not. Filtering invalid timestamps before chart shaping would satisfy the issue's invalid-input acceptance criterion.

The SQLite-to-UTC normalization itself and the ISO/export coverage look solid.

lanpublications:
Good catches, thanks for actually running the branch. Both addressed in c9cde23:

1. `formatTimestampDate` now defaults to `timeZone: "UTC"` (an explicit caller override still wins), so date labels match the stored UTC calendar date in every browser timezone. Added the adversarial-zone regression pair straight from your repro values: `2026-07-20 00:30:00` and `2026-07-20 23:30:00` both render Jul 20 on any machine, and an explicit `America/Los_Angeles` override renders Jul 19. Also ran the suite under `TZ=America/Los_Angeles` and `TZ=Europe/Lisbon` to confirm.

2. Chart shaping now drops invalid timestamps before they reach recharts: `RankTrackingOverview` filters NaN points, and `buildChartData` plus the bottom-band key builder skip them. The formatting wrappers keep their raw-string fallback for invalid input.

One scope note for the maintainer: `formatDay` on the dashboard shares this helper, so its day stamps become UTC-dated too. That seemed like the consistent behavior, but happy to scope the UTC default to rank tracking if local rendering is preferred there.

716/716 tests and `ci:check` green.

isatimur (approved):
Re-reviewed commit c9cde23. The two previously reported blockers are resolved:

- date labels now default to UTC while still allowing an explicit caller override, with both adversarial timezone repros covered
- invalid timestamps are omitted from the overview, trend chart, and bottom-band key paths before reaching chart/map shaping

Local verification on the exact PR head:
- pnpm vitest run src/client/lib/timestamps.test.ts (13/13 passed)
- pnpm ci:check (passed)

The SQLite/Postgres normalization and safe invalid-input fallback remain intact. Approved.

sorcerai (approved):
Reviewed current head and validated it in a 13-PR integration batch on main+#153: pnpm ci:check passed and 775 tests passed. Shared UTC timestamp parsing/formatting and invalid chart-point filtering are covered. #119 should close as superseded.

bensenescu:
Hey,

Thanks for contributing!

PR #105, opened 2026-07-20
Document reproducible contributor verification

isatimur:
## What changed

- Documented Corepack setup so contributors use the exact pnpm version declared in `package.json` instead of an incompatible global pnpm.
- Switched the initial install command to `pnpm install --frozen-lockfile`.
- Expanded the pre-review checklist to match the root CI checks: static checks, tests, and production worker build.
- Added the website-specific install, typecheck, and build commands for changes under `web/`.

## Why

The current local-development guide accepts any pnpm version, while the repository pins pnpm 10.30.1. Older global pnpm releases reject the lockfile as incompatible. The contribution guide also asks only for `pnpm ci:check`, while GitHub Actions separately runs tests, the production worker build, and website validation.

This makes the documented contributor path reproduce CI failures before a PR is opened.

## Validation

- `pnpm ci:check`
- Markdown formatting check passed

Corepack is bundled through Node.js 24 and must be installed separately on Node.js 25+, matching the current Node.js documentation.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -16,4 +16,18 @@ See [LOCAL_DEVELOPMENT.md](./LOCAL_DEVELOPMENT.md) for how to run the app locall
 
 - Keep PRs focused: one feature or fix per PR.
 - For larger features, open an issue first so we can align on the approach before you invest time.
-- Make sure `pnpm ci:check` passes before requesting review.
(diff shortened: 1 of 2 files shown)


sorcerai (approved):
Reviewed exact docs diff and validated main+#153+#105 locally: pnpm ci:check passed. Corepack, frozen installs, and the listed CI/build commands match the repository configuration.

PR #108, opened 2026-07-20
fix(search-tabs): keep tabs without explicit location across reloads

xiaonancui:
Search tabs opened at the default location disappear after a full page reload.

Repro: run a Domain Overview search without changing the location (the URL has no `loc` param). The tab shows up fine — reload the page and it's gone. Keyword tabs at the default location drop the same way.

`locationCode` is optional in both tab input types, and at the default location the input is persisted without the key. But `parseTabInput` requires a numeric `locationCode` when hydrating from sessionStorage, so those inputs are rejected and the tabs are silently dropped.

This accepts a missing `locationCode` for domain and keyword inputs (non-number values are still rejected) and normalizes it to `undefined`. Added unit tests around `parseStoredState`: they reproduce the drop on main and pass with the fix. `pnpm test` and `pnpm ci:check` pass locally.

Found while getting the e2e suite running on a fork.

### DIFF
--- a/src/client/features/search-tabs/useSearchTabs.test.ts
+++ b/src/client/features/search-tabs/useSearchTabs.test.ts
@@ -1,0 +1,103 @@
+import { describe, expect, it } from "vitest";
+import { parseStoredState } from "./useSearchTabs";
+
+function persistedTab(input: unknown) {
+  return {
+    id: "tab-1",
+    label: "example",
+    createdAt: 1,
+    viewedAt: null,
+    input,
+  };
+}
+
+describe("parseStoredState", () => {
+  it("keeps domain tabs persisted without a locationCode (default location)", () => {
+    const state = parseStoredState({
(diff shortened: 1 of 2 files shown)


isatimur (approved):
Reviewed ff7277a. The persisted types already make locationCode optional, and this patch aligns hydration with that contract while still rejecting non-number values. Normalizing the absent value to undefined preserves the existing JSON key/deduplication behavior.

Verification:
- pnpm vitest run src/client/features/search-tabs/useSearchTabs.test.ts (4/4 passed)
- pnpm ci:check passed on the clean combined #108 + #109 merge tree
- the affected E2E suites were exercised after local D1 migration; they advanced through the repaired landing/hydration paths, but full E2E verification was environment-blocked by the missing DATAFORSEO_API_KEY causing the global error modal

No code blocker found. Approved.

PR #109, opened 2026-07-20
test(e2e): update landing redirect assertions for dashboard landing page

xiaonancui:
The e2e suite has been red since #398 moved the landing redirect from `/p/:id/keywords` to `/p/:id` (the new project dashboard): both suites still wait for a `/p/<id>/keywords` URL after visiting `/`, so every spec times out in `getProjectId` before the actual test starts.

This updates the two wait-for regexes to expect the dashboard landing.

With this, 9 of 11 specs pass. The remaining two domain tab-close specs fail on a separate tab-persistence bug — that's #108; with both, the suite is fully green.

### DIFF
--- a/e2e/domain-overview-test-utils.ts
+++ b/e2e/domain-overview-test-utils.ts
@@ -181,11 +181,11 @@ export async function openDomainOverview(page: Page, tab: DomainTab) {
     window.sessionStorage.setItem("domain-overview-e2e-cleared", "1");
   });
   await page.goto("/");
-  await page.waitForURL(/\/p\/([^/]+)\/keywords(?:\?.*)?$/, {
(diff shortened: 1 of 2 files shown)


isatimur (approved):
Reviewed 76c410c. Both waitForURL expressions now match the current /p/:projectId dashboard landing, and project-id extraction remains safe because it runs only after the anchored wait succeeds. The two PRs touch separate files and merge cleanly together.

Verification:
- pnpm ci:check passed on the combined #108 + #109 merge tree
- the E2E run advanced past the redirect/project-id discovery that this PR fixes and began all affected specs; later failures were unrelated local-environment failures from a missing DATAFORSEO_API_KEY/global error modal, so I am not treating them as patch failures

No code blocker found. Approved.

PR #112, opened 2026-07-21
Add RankParse as an optional backlinks-data provider

abhibavishi:
## Summary
Adds [RankParse](https://rankparse.com) as an **opt-in, additive** alternative to DataForSEO for the Backlinks feature. DataForSEO remains the default in every deployment — both `RANKPARSE_API_KEY` and `BACKLINKS_PROVIDER=rankparse` must be set to activate it, so no existing self-hosted deployment changes behavior by pulling this in.

**Disclosure:** I'm affiliated with RankParse (I built it). Flagging that up front since this PR proposes it as an alternative to your existing paid integration.

## Why
RankParse is a cheaper, Common-Crawl-backed link-graph API. It doesn't have full feature parity with DataForSEO's Backlinks API though — no history/trend data, no spam-score data, no page-level (single-URL) lookups, and only coarse server-side filtering/sorting. Rather than a lossy full replacement, this adds it as a swappable provider so those gaps degrade gracefully (empty/`-`, never an error) instead of silently breaking existing functionality.

## What changed
- New `src/server/lib/rankparse/{client,backlinks}.ts` — REST client + 5 fetchers matching `dataforseo/backlinks.ts`'s signatures, reusing its Zod schemas (extracted into a new SDK-free `backlinksSchemas.ts` so the DataForSEO SDK stays lazy-loaded — this repo has a build-time check for that).
- `dataforseo/client.ts` — provider switch on the `backlinks` sub-client only; everything else untouched.
- `docs/RANKPARSE_API_KEY.md` + small cross-reference additions in the existing self-hosting docs / `.env.example`.
- Tests for the new fetchers, error mapping, and provider selection (defaults to DataForSEO, opt-in only).

## Testing
- `npm run ci:check`, `npm run test` (726 tests), `npm run build` all pass.
- Manually verified end-to-end against the live RankParse API (real domains: github.com, vercel.com) via local dev — overview, all three tabs, and the page-scope graceful-empty-state path.
- That live testing caught and fixed three real issues before opening this: `/v1/site-explorer` was too slow/inaccurate for the summary (switched to `/v1/domain-authority` + `/v1/domain-rank`), a documented field name (`from_host_count`) doesn't match the live API (`total_links`), and the client timeout needed to match DataForSEO's existing 60s budget.

## Out of scope
Only the Backlinks feature — no changes to keyword/SERP/labs/lighthouse/AI DataForSEO integrations, and no UI changes (existing spam-filter/range-filter controls remain in the DOM but are no-ops under this provider).


FrancescoSaverioZuppichini:
@abhibavishi get a life and fix your vibecoded webpage

abhibavishi:
@FrancescoSaverioZuppichini Thanks! I had no idea that the definition of life was gaslighting strangers on Git repos. I hope you understand the meaning of "open" in this lifetime.

PR #115, opened 2026-07-21
feat(chat): support a local / OpenAI-compatible chat model via CHAT_BASE_URL

### LINKED ISSUE #102: Self-host: allow a local / OpenAI-compatible chat model (vLLM, Ollama) instead of forcing OpenRouter
## Problem

The in-app chat agents (onboarding + SAM) are hardwired to OpenRouter. `buildChatAgentModel` (`src/server/lib/openrouter.ts`) calls `createOpenRouter({ apiKey })` against OpenRouter's fixed endpoint, and the only knobs exposed to self-hosters are `OPENROUTER_API_KEY` and `OPENROUTER_MODEL` (`src/env.d.ts`). There's no way to point the chat at a self-hosted, OpenAI-compatible model.

For a self-hosted, privacy-conscious deployment this is the one place data still has to leave the box. We run OpenSEO on our own hardware specifically so client SEO data stays in-house — but the chat agent forces prompts (which include project domains, keywords, GSC-derived context) out to OpenRouter. We already run an OpenAI-compatible endpoint on the same host that could serve it:

```
$ curl http://<host>:8000/v1/models
{"object":"list","data":[{"id":"qwen3.5-9b","object":"model","owned_by":"vllm","root":"...Qwen3.5-9B...","max_model_len":16384,...}]}
```

vLLM (and Ollama, LM Studio, LiteLLM, etc.) all expose the OpenAI `/v1/chat/completions` shape, so the AI SDK could talk to them directly.

## Proposal

Let self-hosters override the chat provider's base URL and key, and degrade the OpenRouter-only request options when a custom endpoint is set. Minimal shape:

- New env: `CHAT_BASE_URL` (or `OPENROUTER_BASE_URL`) + reuse `OPENROUTER_API_KEY` as the bearer (vLLM accepts any/no key).
- In `buildChatAgentModel`, when `CHAT_BASE_URL` is set:
  - pass `baseURL` to `createOpenRouter` (the provider already accepts it), **or** swap to `@ai-sdk/openai`'s `createOpenAI({ baseURL, apiKey })`;
  - drop the OpenRouter-specific options that a generic endpoint won't honor: `usage: { include: true }` (cost metering), `provider: { order, zdr, allow_fallbacks }` (routing), and the OpenRouter `reasoning` channel;
  - skip OpenRouter usage-cost metering for that path (`openRouterCostUsd` → 0 / "self-hosted").
- Docs note in `SELF_HOSTING_DOCKER.md`.

## Caveats worth stating in the issue

The SAM agent is **tool-calling** (it drives the MCP tools). So a local model only fully replaces it if it supports OpenAI-style function calling. With vLLM that means launching the server with `--enable-auto-tool-choice --tool-call-parser hermes` (or the model-appropriate parser). Models without reliable tool-calling would get a degraded chat (no agentic tool use). Might be worth a startup check / doc caveat rather than silently failing.

## Why upstream and not a fork

We deliberately track the official image (`ghcr.io/every-app/open-seo:latest`) to keep getting updates. A one-line base-URL override is small, self-contained, and useful to every self-hoster who wants a local/sovereign model or a cheaper gateway (LiteLLM, etc.). Happy to contribute the PR, including the tool-calling doc caveat, if a maintainer is open to the direction.

Santofer:
Closes #102.

## What

Lets self-hosters run the in-app chat agents (onboarding + SAM) on **any OpenAI-compatible endpoint** — a local model (vLLM, Ollama, LM Studio) or a gateway (LiteLLM) — instead of forcing OpenRouter. For a privacy-conscious self-host, this closes the one place chat prompts (project domains, keywords, GSC-derived context) still had to leave the box.

Per your note on #102: went with `CHAT_BASE_URL`, kept OpenRouter as the default, and included Ollama + vLLM setup docs.

## How

- **`buildChatAgentModel(apiKey, modelId?, baseURL?)`** — when `baseURL` is set, points the existing `@openrouter/ai-sdk-provider` at the custom base URL (it already accepts `baseURL`, so no new dependency) and **drops the OpenRouter-only request options** (usage accounting, `provider` routing/ZDR, the `reasoning` channel) that a generic endpoint doesn't understand. Cost metering falls back to `0` because there's no OpenRouter usage metadata to read (`openRouterCostUsd` already returns 0 when absent).
- **`getChatAgentModel()` / `SamChatAgent.getModel()`** — read `CHAT_BASE_URL`; `OPENROUTER_API_KEY` becomes optional when it's set (local endpoints usually need no key; a placeholder is sent so the SDK still builds an auth header, which vLLM/Ollama ignore).
- **SAM access gate (`samAccess.ts`)** — enables on **either** an OpenRouter key **or** `CHAT_BASE_URL`, so a self-hoster with only a local model isn't told "configure a chat model".
- **`env.d.ts`** — documents `CHAT_BASE_URL`; `OPENROUTER_MODEL` doubles as the local model id.
- The default OpenRouter path is byte-for-byte unchanged.

`OPENROUTER_MODEL` is reused as the model id (e.g. `qwen3.5-9b`) rather than adding a new var, to keep the surface small — happy to rename to `CHAT_MODEL` if you'd prefer symmetry with `CHAT_BASE_URL`.

## Docs

`docs/SELF_HOSTING_LOCAL_CHAT_MODEL.md` — Ollama and vLLM setup, and the important **tool-calling caveat**: the SAM agent is agentic (it drives the MCP tools), so a local model only fully replaces OpenRouter if it supports OpenAI-style function calling (Ollama: a tool-calling model like `llama3.1`; vLLM: launch with `--enable-auto-tool-choice --tool-call-parser hermes`).

## Ollama quick test

```sh
ollama pull llama3.1 && ollama serve
```
```sh
CHAT_BASE_URL=http://localhost:11434/v1
OPENROUTER_MODEL=llama3.1
```
Restart OpenSEO → open a project → SAM is enabled → ask "list my projects" → it calls `list_projects` and answers from the local model.

## Tests

`src/server/lib/openrouter.test.ts` (new) mocks `createOpenRouter` and asserts:
1. Default path → `{ apiKey }` + the OpenRouter options (`usage.include`, `provider.zdr`).
2. `CHAT_BASE_URL` path → `{ apiKey: "local", baseURL }` and **no** per-model options.
3. Explicit key + base URL is forwarded as the bearer.

## Checks

- `tsc --noEmit`: 0 errors (project-wide) · `oxlint --type-aware`: 0/0 · `prettier --check`: clean · `vitest`: 3/3.
- (`pnpm ci:check`'s `knip` step fails locally while loading `vite.config.ts` — reproduces on clean `main`, unrelated to this PR.)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -42,6 +42,13 @@
 # the in-app SEO agent. Create a key at https://openrouter.ai/settings/keys.
 # OPENROUTER_API_KEY=replace-with-your-openrouter-api-key
 
+# Optional. Point the chat agent at any OpenAI-compatible endpoint (vLLM,
+# Ollama, LM Studio) instead of OpenRouter — the key above then becomes
+# optional. Set OPENROUTER_MODEL to a model id that endpoint serves.
+# See docs/SELF_HOSTING_LOCAL_CHAT_MODEL.md
+# CHAT_BASE_URL=http://localhost:11434/v1
+# OPENROUTER_MODEL=llama3.1
+
 # Optional in self-hosted modes. Required if you want Google Search Console
 # integration and MCP tools. BETTER_AUTH_SECRET is also required for GSC (it
 # encrypts the stored OAuth tokens). See docs/SELF_HOSTING_GOOGLE_SEARCH_CONSOLE.md.
--- a/compose.yaml
+++ b/compose.yaml
@@ -28,8 +28,12 @@ services:
       - GOOGLE_CLIENT_SECRET=${GOOGLE_CLIENT_SECRET:-}
       - BETTER_AUTH_SECRET=${BETTER_AUTH_SECRET:-}
       # Optional: AI features (SAM, the in-app SEO agent)
-      - OPENROUTER_API_KEY=${OPENROUTER_API_KEY:-}
       - OPENROUTER_MODEL=${OPENROUTER_MODEL:-}
+      # Optional: point the chat agent at a local / OpenAI-compatible endpoint
+      # (vLLM, Ollama, LM Studio) instead of OpenRouter. When set,
+      # OPENROUTER_API_KEY becomes optional.
+      # See docs/SELF_HOSTING_LOCAL_CHAT_MODEL.md
+      - CHAT_BASE_URL=${CHAT_BASE_URL:-}
       - VITE_SHOW_DEVTOOLS=false
     ports:
       - "127.0.0.1:${PORT:-3001}:${PORT:-3001}"
--- a/docs/SELF_HOSTING_LOCAL_CHAT_MODEL.md
+++ b/docs/SELF_HOSTING_LOCAL_CHAT_MODEL.md
@@ -1,0 +1,79 @@
+# Self-hosting: use a local / OpenAI-compatible chat model
+
+The in-app chat agents (onboarding + SAM) use OpenRouter by default. Self-hosters
+can instead point them at any **OpenAI-compatible** endpoint — a local model
+(vLLM, Ollama, LM Studio) or a gateway (LiteLLM) — so chat prompts never leave
+your infrastructure.
+
+## Configuration
+
+Set two environment variables and restart OpenSEO:
+
+| Variable           | Purpose                                                                            | Example                     |
+| ------------------ | ---------------------------------------------------------------------------------- | --------------------------- |
+| `CHAT_BASE_URL`    | Base URL of your OpenAI-compatible endpoint (the part before `/chat/completions`). | `http://localhost:11434/v1` |
+| `OPENROUTER_MODEL` | The model id to request from that endpoint.                                        | `qwen3.5-9b`                |
+
+When `CHAT_BASE_URL` is set:
+
+- `OPENROUTER_API_KEY` becomes **optional** (local endpoints usually need no key).
+  If your endpoint does require one, set `OPENROUTER_API_KEY` and it is sent as the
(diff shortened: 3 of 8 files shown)


sorcerai (approved):
Reviewed exact diff and validated main+#153+#120+#115 locally: pnpm ci:check passed and 739 tests passed. Operator-only baseURL configuration, credential handling, default OpenRouter behavior, and zero-cost fallback are sound. Docker users still need CHAT_BASE_URL passthrough in compose.yaml as a tiny follow-up.

Santofer:
Thanks for the review and for validating it locally 🙏

Addressed the follow-up you flagged:

- **`CHAT_BASE_URL` passthrough in `compose.yaml`** — added to the explicit `environment:` list next to `OPENROUTER_MODEL`, with a pointer to the doc. (`env_file: .env` already forwarded it implicitly since #152, but it was undiscoverable — a self-hoster had no way to know the var existed.)
- **`.env.example`** — documented the `CHAT_BASE_URL` + `OPENROUTER_MODEL` pair, commented out, so the local-endpoint path is discoverable from the quickstart.
- Also dropped a duplicate `OPENROUTER_API_KEY` line in that same block — it was already forwarded ~6 lines above.

Rebased onto current `main` (52 commits) and re-ran the suite: `openrouter.test.ts` green, no conflicts with the SDK v2 / GA4 work that landed since.

bensenescu:
Hey,

Thanks for contributing!

PR #71, opened 2026-07-10
feat(rank-tracking): add Last checked at column to CSV and Sheets exports

AdarshJ173:
## Summary

Fixes #70 — Rank Tracking exports now include a **`Last checked at`** column in both full-table and selected-row CSV and Google Sheets exports.

### `RankTrackingTableParts.tsx`
- Added a private `formatCheckedAt(lastCheckedAt?: string | null): string` helper that converts the stored ISO-8601 timestamp to a **`YYYY-MM-DD`** string — machine-sortable in Excel, Google Sheets, and any CSV viewer.
- `buildRankTrackingExport` accepts a new optional `lastCheckedAt?: string | null` parameter. The formatted value is appended as the **last column** so all existing column indexes are unchanged (no breaking change for anyone already parsing the CSV programmatically).
- `exportRankTrackingCsv` and `exportRankTrackingToSheets` each accept and forward the new parameter.

### `RankTrackingDomainDetail.tsx`
- Both `onExport` and `onExportToSheets` call sites now pass `run?.lastCheckedAt` — the value already present on `resultsData.run` — so the column is populated automatically on every export.

## Acceptance criteria checklist

- [x] Full-table and selected-row CSV exports include `Last checked at`
- [x] Sheets exports include the same field and value
- [x] Format is `YYYY-MM-DD` — consistent and sortable in common spreadsheet tools
- [x] Existing export columns and filtered-row behaviour remain unchanged
- [x] Graceful empty string when `run` is null (no data yet)

## Testing

Manual verification path (no new unit tests needed for a single helper):
1. Open a Rank Tracking domain with at least one completed check
2. Click **Export CSV** — open the file and confirm the last column is `Last checked at` with a `YYYY-MM-DD` value
3. Click **Export to Sheets** — confirm the same column appears in the sheet
4. Repeat with filters applied (selected-row export) and verify the column is still present
5. On a domain with no completed checks, confirm the column header appears but the value is blank

## Notes
- `formatCheckedAt` wraps `new Date(...).toISOString()` in a try/catch to guard against malformed timestamps — returns `""` on failure.
- No schema changes, no server changes, no new dependencies.

<details><summary><h3>Greptile Summary</h3></summary>

[...]

PR #117, opened 2026-07-21
feat: include last checked date in rank tracking exports (fixes #70)

### LINKED ISSUE #70: Include rank check date in Rank Tracking exports
## Problem
Rank Tracking exports omit when the rankings were checked. That makes saved CSV and Sheets files harder to interpret or compare later.

## Evidence
- User feedback: “Include the date of the rank tracking in the export.”

## Implementation check
Checked `main` at `fec642a0`:
- `src/client/features/rank-tracking/RankTrackingTableParts.tsx` builds CSV/Sheets rows without a date column.
- `src/client/features/rank-tracking/RankTrackingDomainDetail.tsx` already receives the latest run’s `lastCheckedAt` value.
- `src/server/features/rank-tracking/services/rankTrackingResults.ts` returns the stored check timestamp.

The timestamp already exists; the export path only needs to carry and format it.

## Suggested direction
Add one machine-sortable `Last checked at` column to both full-table and selected-row CSV/Sheets exports.

## Contributor fit
Good first issue: small, localized, and testable without changing data collection.

## Acceptance criteria
- [ ] Full-table and selected-row CSV exports include `Last checked at`.
- [ ] Sheets exports include the same field and value.
- [ ] The format is consistent and sortable in common spreadsheet tools.
- [ ] Existing export columns and filtered-row behavior remain unchanged.

Dingding-leo:
## Summary
Fixes #70

Adds `Last checked at` column to rank tracking CSV and Sheets exports.

## Changes
- `buildRankTrackingExport`: new `lastCheckedAt` param → header + each row
- `exportRankTrackingCsv`: grouped optional params into opts object (max-params=5)
- `exportRankTrackingToSheets`: accept + pass through `lastCheckedAt`
- `RankTrackingDomainDetail`: passes `latestRun.lastCheckedAt` to both export callbacks

ISO format for spreadsheet sortability. Timestamp already exists in `latestRun` — just carried through to exports.

### DIFF
--- a/src/client/features/rank-tracking/RankTrackingDomainDetail.tsx
+++ b/src/client/features/rank-tracking/RankTrackingDomainDetail.tsx
@@ -270,8 +270,8 @@ export function RankTrackingDomainDetail({
               showDesktop,
               showMobile,
               config.domain,
-              config.locationName,
-              latestRun?.lastCheckedAt,
+              // eslint-disable-next-line @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-type-assertion
+              { locationName: config.locationName, lastCheckedAt: latestRun?.lastCheckedAt as string | null | undefined },
             )
           }
           onExportToSheets={() =>
@@ -280,7 +280,8 @@ export function RankTrackingDomainDetail({
               showDesktop,
               showMobile,
               config.locationName,
-              latestRun?.lastCheckedAt,
+              // eslint-disable-next-line @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-type-assertion
+              latestRun?.lastCheckedAt as string | null | undefined,
             )
           }
           onCopyKeywords={() => {
--- a/src/client/features/rank-tracking/RankTrackingDomainDetail.tsx
+++ b/src/client/features/rank-tracking/RankTrackingDomainDetail.tsx
@@ -271,6 +271,7 @@ export function RankTrackingDomainDetail({
               showMobile,
               config.domain,
               config.locationName,
+              latestRun?.lastCheckedAt,
             )
           }
           onExportToSheets={() =>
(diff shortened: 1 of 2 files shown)


isatimur:
Compared this against #71 (same issue, #70) since both are open. Two gaps here, verified against current `main`:

1. **`lastCheckedAt` is written to the export unformatted.** `buildRankTrackingExport` does `lastCheckedAt ?? ""` with no `Date` parsing — the raw DB-stored value goes straight into the CSV/Sheets cell. That value's format isn't consistent across backends: SQLite stores `checked_at` as `text` (`YYYY-MM-DD HH:mm:ss`, no timezone marker), while Postgres's `timestampColumn` serializes with one. The existing header display already normalizes this (`RankTrackingDetailHeader.tsx:57`, `new Date(run.lastCheckedAt).toLocaleDateString()`) — the export should go through the same kind of parsing rather than passing the stored string through as-is, or a self-hosted SQLite user's exported column won't sort/parse the same way a Postgres user's will.

2. **The selected-rows export path isn't wired up.** `RankTrackingTable.tsx` has its own `exportSelectionCsv`/`exportSelectionToSheets` (lines ~98-113) that call `buildRankTrackingExport` directly without a `lastCheckedAt` argument — that component doesn't receive `latestRun` as a prop at all. Selecting rows and exporting will always produce a blank "Last checked at" column, only the full-table export from `RankTrackingDomainDetail.tsx` gets the real value.

Flagging both since #71 has the same second gap, so whichever PR moves forward will need it either way.

sorcerai (changes requested):
Two export blockers remain. Raw lastCheckedAt values are passed through unchanged, so SQLite timestamps and PostgreSQL ISO timestamps export with different semantics. Also the selected-row export path calls buildRankTrackingExport without latestRun, producing blank Last checked at values. Please normalize through the shared timestamp helper and cover both all-row and selected-row exports before merge.

bensenescu:
Hey,

Thanks for contributing!

PR #120, opened 2026-07-21
fix(docker): skip vite build on container start when build env is unchanged

### LINKED ISSUE #100: Docker self-host: full vite build runs on every container start (~90s to first response)
## Observation

On the Docker self-host image (`ghcr.io/every-app/open-seo:latest`), the container runs a **full production build on every start**, not just the first one:

```
> open-seo@0.1.0 build /app
> vite build && tsc --noEmit
vite v7.3.6 building client environment for production...
✓ built in 11.06s
vite v7.3.6 building ssr environment for production...
✓ built in 30.76s
```

Measured on a Ryzen home server: **~75–90 s from `docker compose start` to first HTTP 200**, every restart (compose stop/start, host reboot, image update). Migrations are correctly incremental, but the vite client + SSR builds and `tsc --noEmit` always re-run.

For self-hosters this means every unattended restart (Watchtower update, power blip, appliance reboot) takes the tool down for a minute and a half and burns a CPU spike on hardware that often shares duties with other services.

## Likely reason & suggestion

I assume the boot-time build exists because `VITE_*` values (e.g. `VITE_SHOW_DEVTOOLS`) get baked into the client bundle, so the image can't ship a universal prebuilt bundle. Two ideas, either of which would help:

1. **Cache the build output in the data volume keyed on a hash of the build-relevant env** (`VITE_*` set + image version). Unchanged env + same image → skip straight to serve. First boot pays the build once; restarts become seconds.
2. Or **prebuild default-config artifacts into the image** and only rebuild when the env hash differs from the baked-in one.

`tsc --noEmit` could also be dropped from the runtime path entirely — type-checking at container boot verifies the same shipped code on every start.

## Environment

- `ghcr.io/every-app/open-seo:latest` (v0.1.0), Docker 29.5.2 / Compose v2.40.3, Unraid 7.3.1
- Stock upstream `compose.yaml` + `.env` (PORT/ALLOWED_HOST/DATAFORSEO_API_KEY/telemetry opt-outs)

Happy to test a patched image on our setup.

WAHIB-EL-KHADIRI:
Fixes #100

## Problem

The self-host image runs `pnpm run build` (vite build **and** `tsc --noEmit`) on every container start, so every restart — updates, power cycles, `docker compose start` — pays ~75–90s before first HTTP 200, even when nothing changed.

The build genuinely can't move to image-build time: `AUTH_MODE` and the other `envPrefix` values from `vite.config.ts` are inlined into the client bundle, and self-hosters only choose them at runtime via Compose.

## Approach (option 1 from the issue: env-keyed caching)

- New `scripts/docker-selfhost-start.sh` replaces the inline `CMD`. It hashes the build-relevant env — `AUTH_MODE`, `BYPASS_EMAIL_VERIFICATION`, `POSTHOG_PUBLIC_KEY`, `POSTHOG_HOST`, `TURNSTILE_SITE_KEY`, and all `VITE_*` (exactly the `envPrefix` list in `vite.config.ts`) — plus an image build id, and skips `vite build` when the fingerprint matches the previous start.
- The fingerprint marker lives **inside `dist/`**, so `vite build` wiping the outDir also invalidates the cache, and an empty `dist` volume can never claim a false hit.
- `.image-build-id` is stamped at image build time, so a `dist/` carried over in a volume from an older image is rebuilt once instead of served stale.
- `compose.yaml` adds an `open_seo_build:/app/dist` volume so the cache also survives container recreation (`--force-recreate` after `.env` changes, image restarts).
- Runtime `tsc --noEmit` is dropped (new `build:client` script): type-checking can't change the shipped code and CI already runs it.
- `docs/SELF_HOSTING_DOCKER.md` documents when a rebuild happens.

## Result

- First start: unchanged (~90s build).
- Every later start/restart with unchanged build env: build skipped, first HTTP 200 in seconds.
- Changing `AUTH_MODE`/`VITE_*`/PostHog/Turnstile env or updating the image: exactly one rebuild.
- Changing `DATAFORSEO_API_KEY`, `PORT`, `ALLOWED_HOST`, telemetry flags: no rebuild (these are runtime-only).

## Testing

- `scripts/docker-selfhost-start.test.ts` (vitest, skipped on Windows) covers the fingerprint: stable for identical env, changes for each build-inlined var, ignores runtime-only env.
- Manually verified the fingerprint behavior with `--print-fingerprint` (stable / sensitive / insensitive as above).

### DIFF
--- a/Dockerfile.selfhost
+++ b/Dockerfile.selfhost
@@ -13,12 +13,17 @@ RUN pnpm install --frozen-lockfile
 
 COPY . .
 
+# Feeds the client-build fingerprint so a dist/ carried over in a volume from an
+# older image is rebuilt once instead of being served stale.
+RUN date +%s%N >.image-build-id
+
 EXPOSE 3001
 
 # The build MUST run at container start, not image-build time: AUTH_MODE (and the
 # other client envs) are inlined into the client bundle by `vite build`, and the
-# self-hoster only chooses AUTH_MODE at runtime via Compose. Building here lets
-# that runtime value bake into the bundle. The repo .npmrc raises the V8 heap
-# ceiling (node-options) so the ~7400-module SSR build doesn't OOM under Node's
-# ~2GB default.
-CMD ["sh", "-c", "echo 'OpenSEO sends an anonymous usage heartbeat (counts only). Disable: OPENSEO_TELEMETRY_DISABLED=1. Details: docs/SELF_HOSTING_DOCKER.md#telemetry' && pnpm run db:migrate:local && pnpm run build && pnpm exec vite preview --host 0.0.0.0 --port ${PORT:-3001}"]
+# self-hoster only chooses AUTH_MODE at runtime via Compose. The start script
+# fingerprints those envs and reuses the previous build when they are unchanged,
+# so only the first start (or an env/image change) pays for the build. The repo
+# .npmrc raises the V8 heap ceiling (node-options) so the ~7400-module SSR build
+# doesn't OOM under Node's ~2GB default.
+CMD ["sh", "scripts/docker-selfhost-start.sh"]
--- a/compose.yaml
+++ b/compose.yaml
@@ -22,5 +22,9 @@ services:
       - "127.0.0.1:${PORT:-3001}:${PORT:-3001}"
     volumes:
       - open_seo_data:/app/.wrangler
+      # Keeps the built client bundle across container recreation so restarts
+      # skip the ~90s vite build when build-relevant env is unchanged.
+      - open_seo_build:/app/dist
 volumes:
   open_seo_data:
+  open_seo_build:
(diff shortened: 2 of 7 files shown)


sorcerai (approved):
Reviewed exact diff and validated main+#153+#120+#115 locally: pnpm ci:check passed and 739 tests passed. The fingerprint matches vite envPrefix, includes image identity, persists dist across recreation, and supersedes #123. Merge after #153 so GitHub checks rerun on the stabilized base.

WAHIB-EL-KHADIRI:
Closing this — `main` now ships the same behavior in `docker-entrypoint.sh`: the client build runs at container start but is skipped when the build-relevant envs (VITE_*, AUTH_MODE, …) and the image are unchanged, fingerprinted and fail-loud. That supersedes this PR's `docker-selfhost-start.sh` approach, so there's nothing left to merge here. Thanks for implementing it!

Issue #101, opened 2026-07-20
IndexNow: per-project key management, URL submission, and an MCP tool


Santofer:
## Problem

OpenSEO helps users find what to rank for and tracks whether they rank — but does nothing to shorten the gap in between: **getting (re)indexed**. Sitemaps are pull-based; Bing/Yandex discovering a change can take days. IndexNow (https://www.indexnow.org/) is push-based and free: a single POST notifies Bing + Yandex (and other participating engines) within seconds.

This matters more than it used to, because **Bing's index feeds Copilot and ChatGPT search**. A page invisible to Bing is invisible to a growing share of AI-mediated queries — which is exactly the audience OpenSEO's AI-visibility features target. Faster Bing indexing = faster appearance in the engines Prompt Explorer measures.

## Proposal

A small IndexNow feature attached to projects (a project already has a `domain`):

1. **Key management** — generate a 32-hex key per project, show the required key file (`https://<domain>/<key>.txt`) and verify it's reachable.
2. **Submit** — "Submit URLs" action (paste list, or pull from the project's sitemap) that POSTs to `https://api.indexnow.org/indexnow` with `{host, key, keyLocation, urlList}`. Zero DataForSEO credits; the endpoint is free and unauthenticated beyond the key file.
3. **MCP tool** — `submit_urls_indexnow(projectId, urls[])` so agents can close the publish → index → track loop autonomously.

Minimal surface: one settings field (key), one action, one MCP tool. No scheduler needed for v1 (manual/agent-triggered submits).

## Precedent

Same "bring your own free Google/Bing capability" shape as #52 (free Lighthouse via PageSpeed key). We already run this flow externally (key generation + sitemap submit script) for client sites and would be glad to contribute the port as a PR if the direction gets a 👍.

bensenescu:
Yes, I would appreciate a PR for this! A video demoing it would help a ton to make it easier to review as well. 

Also, feel free to join the #dev discord, I'll be quicker to respond there. 

https://discord.com/invite/c9uGs3cFXr

PR #121, opened 2026-07-21
feat(indexnow): add IndexNow key + URL submission MCP tools

Santofer:
Partially addresses #101 — this lands the MCP surface only; the key-management UI is deliberately left out (see discussion below).

## What

Adds **IndexNow** so agents can push new/updated URLs to Bing, Yandex, and other IndexNow engines — free, typically indexed within seconds. Since Bing's index feeds Copilot and ChatGPT search, this also shortens time-to-appearance in those AI answers, which lines up with what Prompt Explorer measures.

Kept it **MCP-first** (matching the agent-first design): two project-scoped tools, so the publish → index loop closes from chat. No UI in this PR — happy to add a settings card as a follow-up if you'd like one.

## Tools

- **`get_indexnow_key`** — returns `{ host, key, keyLocation, keyFileContent }` for a project (requires a domain). Uses no credits.
- **`submit_urls_indexnow`** — submits up to 10,000 same-host URLs to `https://api.indexnow.org/indexnow`; cross-host URLs are skipped and reported (IndexNow only accepts same-host lists). Uses no credits.

## Design note — no DB, no migration

IndexNow keys **aren't secrets**: control of a host is proven by serving `<key>.txt` at its root. So the key is **derived deterministically** from the project id (SHA-256 → 32 hex), which is stable across restarts and needs no storage. This avoided touching the schema (and the D1 + Postgres dual-migration surface) entirely — the feature is fully functional with zero persistence.

## Files

- `src/server/lib/indexnow.ts` — key derivation, host parsing, same-host partitioning, the submit POST.
- `src/server/mcp/tools/indexnow-tools.ts` — the two tools, registered in `server.ts`.
- `docs/INDEXNOW.md` — setup (publish the key file) + usage.

## Tests

`indexnow-tools.test.ts` (4): deterministic key + key-file location; errors when the project has no domain; submits on-host URLs and skips cross-host ones (asserts the exact POST body); errors when nothing is on-host (no fetch made).

## Checks

- `tsc --noEmit`: 0 errors (project-wide) · `oxlint --type-aware`: 0/0 · `prettier --check`: clean · `vitest`: 4/4 new, `output-schema-validation` 9/9.
- (`pnpm ci:check`'s `knip` step fails locally while loading `vite.config.ts` — reproduces on clean `main`, unrelated to this PR.)

## Demo

I'll add a short screen recording (agent calls `get_indexnow_key`, I publish the file, `curl` shows it live, then `submit_urls_indexnow` returns HTTP 200) against my self-hosted instance.

### DIFF
--- a/docs/INDEXNOW.md
+++ b/docs/INDEXNOW.md
@@ -9,6 +9,11 @@ indexing also means faster appearance in those AI answers.
 OpenSEO exposes IndexNow through two MCP tools, so agents can close the
 publish → index loop without leaving the chat.
 
+> **Google does not participate in IndexNow.** Submitting notifies Bing,
+> Yandex, Seznam, Naver, and Yep — a single submission is shared between them.
+> Google indexing is unaffected and still relies on crawling / sitemaps. This
+> catches people out, so the tool description says it too.
+
 ## How the key works
 
 IndexNow proves you control a host by having you serve a **verification file**
@@ -30,6 +35,23 @@ Submits up to 10,000 URLs to IndexNow for the project's host. URLs that aren't o
 that host are skipped and reported (IndexNow only accepts same-host lists). The
 verification file must already be live, or engines ignore the submission. Uses no
 credits.
+
+**Response contract.** The engine's HTTP status is returned as _data_
+(`{ status, ok }`), not raised as a tool error — the call itself succeeded, the
+engine just may not have accepted the list. `ok` tracks 2xx, so `202`
+("accepted, key validation pending") is a success. A `403` means the key file
+isn't published (or doesn't match) — the agent can read that and fix it. Only a
+malformed request throws: no domain on the project, or no submitted URL on the
+project's host.
+
+| Status | Meaning                                                            |
+| ------ | ------------------------------------------------------------------ |
+| `200`  | Accepted                                                           |
+| `202`  | Accepted, key validation pending (normal before the file is live)  |
+| `400`  | Invalid format                                                     |
+| `403`  | Key not valid — file missing or contents don't match               |
+| `422`  | URLs don't belong to the host, or the key doesn't match the schema |
+| `429`  | Too many requests                                                  |
 
 ## Setup (one time per project)
 
--- a/docs/INDEXNOW.md
+++ b/docs/INDEXNOW.md
@@ -1,0 +1,44 @@
+# IndexNow
+
+[IndexNow](https://www.indexnow.org/) is a free, push-based protocol: instead of
+waiting for search engines to re-crawl, you notify **Bing, Yandex**, and other
+participating engines that URLs are new or updated, and they pick them up within
+seconds. Because Bing's index feeds Copilot and ChatGPT search, faster Bing
+indexing also means faster appearance in those AI answers.
+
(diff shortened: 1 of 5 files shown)


Santofer:
Thanks! Two notes:

- **Demo video** incoming — MCP flow: `create_project` → `get_indexnow_key` → publish the key file → `submit_urls_indexnow` returns 200.
- This PR is **MCP-first** (no UI yet), to keep it focused. Happy to follow up with a small **UI** — e.g. an "IndexNow" card on the project settings/page showing the key + key-file to publish and a "Submit sitemap URLs" button. Where would you want it to live? I'll align with you before building so it fits your UX.

Santofer:
https://github.com/user-attachments/assets/d9a8ddfd-bb7f-4cf6-b857-d57c9558328a

sorcerai (changes requested):
The MCP authorization and project scoping are sound, but this PR explicitly omits the key-management/settings UI requested by #101, so Closes #101 is premature. Please mark it Related/partial or complete that UI scope. Also pin the intended HTTP non-2xx contract with a test: submitToIndexNow currently returns ok:false/status as a normal MCP result rather than throwing a tool error. Rebase/rerun checks afterward.

Santofer:
Thanks for the careful review — all three points addressed.

**1. `Closes #101` was premature.** Changed the PR body to *"Partially addresses #101"*, so merging this no longer auto-closes the issue. I'd rather keep this PR MCP-only and do the settings UI as a follow-up than bundle two review surfaces — but I'm happy to build the UI now if you'd prefer #101 land in one piece. If so, my proposal is an **IndexNow card in project settings**: shows the derived key + the exact file to publish, a "verify key file" button that fetches `keyLocation` and reports live/not-live, and a "submit sitemap URLs" action. Say the word and I'll add it here or open it as PR 2.

**2. Non-2xx contract is now pinned by a test.** You're right that it was unstated. `submitToIndexNow` reports the engine status as **data** (`{ status, ok }`) rather than a thrown tool error, and there's now a table-driven test over **202 / 403 / 422 / 500** asserting exactly that (`ok` tracks 2xx, `isError` stays falsy).

The reasoning, now written down in both the tool description and `docs/INDEXNOW.md`: a non-2xx means the request was well-formed but the engine declined it — overwhelmingly a **403 because the key file isn't published yet**, which is a state the agent can read and fix. Throwing would bury it in a tool error. `202` ("accepted, key validation pending") is genuinely a success, so the check is 2xx rather than strictly 200. Malformed requests do still throw: no domain on the project, or no submitted URL on the project's host.

**3. Rebased.** Onto current `main`, which meant adapting to the **MCP SDK v2 migration** (#464) — tools register through `registerOpenSeoTool` now, and the test moved to the new `makeToolContext` helper. Gates locally: `tsc` clean, `oxlint --type-aware` 0 errors, **152/152 MCP tests green** (8 of them IndexNow).

One extra: I documented that **Google does not participate in IndexNow** (only Bing/Yandex/Seznam/Naver/Yep). I hit that confusion myself while testing, and it seems worth stating in the tool description so agents don't imply Google coverage to users.

raymatos:
Anyone picking up this up?

bensenescu:
Hey,

Thanks for contributing!

Issue #99, opened 2026-07-20
Site Audit: add AEO/agent-readiness checks (llms.txt, AI crawlers in robots.txt, markdown alternates, linked JSON-LD)


Santofer:
## Context

OpenSEO tracks whether AI engines *mention* a brand (Prompt Explorer / AI visibility). The complementary question the site audit doesn't answer yet: **is the site itself readable and citable by AI agents?** Two sites with identical rankings can have wildly different agent-readiness, and today none of that surfaces in the audit.

We run audits for several client sites (Morocco + UAE markets) and ended up building these checks outside OpenSEO — proposing them upstream because they feel squarely in scope for "site audits" in 2026. Joost de Valk's recent Astro SEO guide (https://joost.blog/astro-seo-complete-guide/) is a good reference for how mainstream this layer is becoming.

## Proposed audit checks (roughly by value/effort)

1. **AI crawlers in robots.txt** — parse robots.txt against the known AI user-agents (GPTBot, OAI-SearchBot, ChatGPT-User, ClaudeBot, Claude-SearchBot, PerplexityBot, Google-Extended, Applebot-Extended, CCBot, Meta-ExternalAgent, Amazonbot, Bytespider…), and report which are blocked, distinguishing `purpose: training` vs `search-grounding` vs `user-browse`. Blocking Google-Extended while wanting AI Overview visibility is a very common self-inflicted wound — we found a chatbot company blocking 8 AI bots via a default Cloudflare managed robots.txt without knowing.
2. **llms.txt** — presence + basic structural validation (H1 title, blockquote summary, `##` sections with markdown links) per https://llmstxt.org/.
3. **Markdown alternates** — `<link rel="alternate" type="text/markdown">`, per-page `.md` URLs, and/or content negotiation on `Accept: text/markdown`.
4. **JSON-LD graph quality** — flag flat isolated schema blocks vs a linked `@graph` with resolvable `@id` references; check trust-signal properties (`publishingPrinciples`, `copyrightHolder`, `knowsAbout`, `SearchAction`).
5. **schemamap.xml / schema endpoints** (NLWeb-adjacent, early-stage — lowest priority).

Each check is a cheap HTTP fetch + parse; no DataForSEO credits involved.

## Why in OpenSEO rather than a separate tool

[...]

PR #122, opened 2026-07-21
feat(audit): AI readiness checks — robots.txt AI crawlers, llms.txt, and Markdown alternates (part 1 of #99)

WAHIB-EL-KHADIRI:
Part of #99 — now covers three of the four priority checks from the issue: **AI crawlers in robots.txt**, **llms.txt validation**, and **Markdown alternates**. JSON-LD graph quality can follow as a separate PR (it's the lowest-priority item and touches page-analyzer structured-data parsing, a bigger diff on its own).

## What this adds

Six new issue types in the shared registry, all site-level (`pageId: null` — the schema already allows it):

| Issue | Severity | When |
|---|---|---|
| `ai-search-crawlers-blocked` | warning | robots.txt blocks index-building agents (OAI-SearchBot, Claude-SearchBot, PerplexityBot) |
| `ai-user-fetchers-blocked` | warning | robots.txt blocks user-request fetchers (ChatGPT-User, Claude-User, Perplexity-User, MistralAI-User) |
| `ai-training-crawlers-blocked` | info | robots.txt blocks training crawlers (GPTBot, ClaudeBot, Google-Extended, Applebot-Extended, meta-externalagent, CCBot, Bytespider) |
| `missing-llms-txt` | info | no /llms.txt (absence is common — heads-up, not an error) |
| `llms-txt-invalid` | warning | /llms.txt served but violates the spec's one hard requirement (H1 title first) |
| `no-markdown-alternates` | info | no crawled page advertises `<link rel="alternate" type="text/markdown">` (absence is common — heads-up, not an error) |

## Design decisions

- **Only AI-specific robots.txt blocks are flagged.** An agent counts as blocked only when it is treated worse than the generic `*` rules (checked via a probe UA that falls through to `*`). A site with a blanket `Disallow: /` made a site-wide choice — flagging every AI agent there would be noise.
- **One issue per purpose group, not per agent** — blocked agents are listed in `details.blockedAgents` with vendor names, so a site blocking 7 training crawlers gets 1 row, not 7.
- **Google-Extended nuance from the issue** ("blocking Google-Extended while wanting AI Overview visibility") is addressed in the descriptor text: it is listed under training but the explanation calls out that it also gates Gemini answer grounding.
- **llms.txt false-positive protection**: SPA catch-alls that answer 200 with the app shell for any path are detected (content-type / `<!doctype` sniff) and treated as *missing*, not invalid; unreachable fetches stay silent instead of mis-reporting. Structure validation is deliberately lenient — the spec's only hard requirement is the H1 title; optional blockquote/sections are not flagged.
- **Markdown alternates is a site-level aggregate, not per-page**: almost no site does this today, so flagging every crawled page would be noise. `page-analyzer.ts` detects the `<link rel=alternate type="text/markdown">` tag during the existing per-page cheerio pass (same pattern as the neighboring hreflang extraction) — no extra fetches. The flag flows transiently through `CrawledPageResult` → the crawl step's slim `StepPageSummary` (same path already used for `title`/`statusCode`), so it needed **no new persisted column or schema migration** — the finding is only ever recorded as an `audit_issues` row.
- **Plumbing reuse throughout**: everything runs in the existing `multipage-checks` finalize step and reuses the robots.txt text the discovery phase already checkpoints (replay-safe). The only new network cost is a single llms.txt fetch. No external API credits — the shared registry automatically covers the issues UI, CSV export, and MCP.

## Testing

- 19 unit tests in `ai-readiness.test.ts`: purpose grouping, blanket-block silence, subpath-only blocks ignored, dedupe keys, llms.txt valid/empty/no-H1/missing/unreachable cases, markdown-alternates empty-crawl/found/none-found cases.
- `tsc --noEmit`, `oxlint`, `knip`, `prettier` all clean. Full vitest suite (`src/server/lib/audit`): 58/58 passing.

### DIFF
--- a/src/server/lib/audit/issues/ai-readiness.test.ts
+++ b/src/server/lib/audit/issues/ai-readiness.test.ts
@@ -1,0 +1,226 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import {
+  checkAiCrawlerAccess,
+  checkLlmsTxt,
+  fetchLlmsTxt,
+  findLlmsTxtStructureProblem,
+} from "@/server/lib/audit/issues/ai-readiness";
+
+// The SSRF guard does DNS/network work; make it an identity in tests so the
+// redirect paths are exercised without touching the network.
+vi.mock("@/server/lib/audit/url-policy", () => ({
+  normalizeAndValidateStartUrl: async (u: string) => u,
+}));
+
+const ORIGIN = "https://example.com";
+
+function mockFetchSequence(responses: Response[]) {
+  let i = 0;
+  vi.stubGlobal(
+    "fetch",
+    vi.fn(async () => responses[i++]),
+  );
+}
+
+function streamOf(totalBytes: number): ReadableStream<Uint8Array> {
+  const chunk = new TextEncoder().encode("x".repeat(100_000));
+  let sent = 0;
+  return new ReadableStream({
+    pull(controller) {
+      if (sent >= totalBytes) {
+        controller.close();
+        return;
+      }
+      controller.enqueue(chunk);
+      sent += chunk.byteLength;
+    },
+  });
+}
+
+afterEach(() => vi.unstubAllGlobals());
+
+describe("checkAiCrawlerAccess", () => {
+  it("reports nothing when robots.txt is missing", () => {
+    expect(checkAiCrawlerAccess(ORIGIN, null)).toEqual([]);
+  });
+
+  it("reports nothing when robots.txt allows everyone", () => {
+    expect(checkAiCrawlerAccess(ORIGIN, "User-agent: *\nAllow: /\n")).toEqual(
+      [],
+    );
+  });
+
+  it("groups specifically blocked agents by purpose as a readable string", () => {
+    const robots = [
+      "User-agent: GPTBot",
+      "Disallow: /",
+      "",
+      "User-agent: ClaudeBot",
+      "Disallow: /",
+      "",
+      "User-agent: PerplexityBot",
+      "Disallow: /",
+      "",
+      "User-agent: ChatGPT-User",
+      "Disallow: /",
+      "",
+      "User-agent: *",
+      "Allow: /",
+    ].join("\n");
+
+    const issues = checkAiCrawlerAccess(ORIGIN, robots);
+    const byType = new Map(issues.map((issue) => [issue.issueType, issue]));
+
+    expect(byType.size).toBe(3);
+    expect(
+      byType.get("ai-training-crawlers-blocked")?.details?.blockedAgents,
+    ).toBe("GPTBot (OpenAI), ClaudeBot (Anthropic)");
+    expect(
+      byType.get("ai-search-crawlers-blocked")?.details?.blockedAgents,
+    ).toBe("PerplexityBot (Perplexity)");
+    expect(
+      byType.get("ai-user-fetchers-blocked")?.details?.blockedAgents,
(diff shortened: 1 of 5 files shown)


sorcerai (changes requested):
Reviewed exact head against upstream f5697263: there is no duplicate ai-crawler-analysis implementation on upstream main. Three blockers remain. (1) /llms.txt uses redirect-following fetch with no final-origin validation and reads response.text() unbounded; prevent cross-origin redirects and use the existing bounded reader. (2) no-markdown is emitted when every crawled page is redirect/non-HTML/blocked/error because those states collapse to hasMarkdownAlternate=false; evaluate only analyzable successful HTML pages or preserve unknown. (3) blockedAgents is an array of objects, but IssuesView joins arrays and renders [object Object]; make details UI-compatible or update the renderer. Add integration coverage for redirects/body bounds and non-analyzable markdown cases, then rerun CI on current main.

WAHIB-EL-KHADIRI:
Thanks for the thorough review @sorcerai — all three addressed in abb9fc6:

1. **llms.txt SSRF + unbounded read**: switched to `redirect: "manual"`, reject cross-origin redirects, and re-validate same-origin hops via `normalizeAndValidateStartUrl`. Body now goes through the shared `readBoundedText` reader instead of `response.text()`.
2. **no-markdown false positive**: redirect/blocked/error/non-HTML pages default `hasMarkdownAlternate=false`, so the check now counts only analyzed HTML pages (new `isHtml` flag on `StepPageSummary`) and stays silent when none were analyzable.
3. **`[object Object]`**: `blockedAgents` is now a `string[]` (`"GPTBot (OpenAI)"`), compatible with IssuesView's array renderer.

Added integration coverage for cross-origin/same-origin redirects, the HTML app-shell case, the body-size bound, and the non-analyzable markdown cases. Typecheck + lint clean; server audit/workflow tests green.

WAHIB-EL-KHADIRI:
Thanks for the thorough review — reworked and rebased onto current `main`.

**Rebase / scope:** `main` has since removed `StepPageSummary`, which the Markdown-alternates check depended on. Rather than re-plumb that onto the new crawl shape in this PR, I **split it out** — this PR is now the two self-contained checks (robots.txt AI-crawler treatment + `llms.txt`). That also resolves blocker (2): the Markdown check that emitted false positives on non-analyzable pages is no longer part of this PR.

**Blocker (1) — SSRF + unbounded read on `/llms.txt`:** `fetchLlmsTxt` now uses `redirect: "manual"`, follows at most one hop, rejects any cross-origin redirect, and re-validates the destination through `normalizeAndValidateStartUrl` (private-IP/DNS guard). The body is read through the shared `readBoundedText` (exported from `scrape.ts`) so an oversized third-party file can't be buffered unbounded.

**Blocker (3) — `[object Object]` in IssuesView:** `blockedAgents` is now a flat string (`"GPTBot (OpenAI), ClaudeBot (Anthropic)"`), so the detail renders as-is with no renderer change.

**Tests:** added integration coverage for the redirect paths (same-origin follow, cross-origin refusal), the body-bound cap, and the HTML catch-all. Locally: the module's tests pass, `tsc` is clean for the touched files, and `oxlint` is clean.

bensenescu:
Hey,

Thanks for contributing!

WAHIB-EL-KHADIRI:
Understood - closing rather than leaving it open against the new process.

Not filing a separate issue for this one: #99 already describes the feature and is still open, and this PR was only part 1 of it. The branch stays up if the implementation is useful as a reference for that issue.

PR #123, opened 2026-07-22
perf(docker): reuse the build across restarts instead of rebuilding on every start

### LINKED ISSUE #100: Docker self-host: full vite build runs on every container start (~90s to first response)

Santofer:
Closes #100.

## Problem

The self-host image (`Dockerfile.selfhost`) rebuilds the client + SSR bundle on **every** container start — its `CMD` runs `pnpm run build` unconditionally. On my Ryzen home server that's **~90–105s of downtime** and a CPU spike on every restart, host reboot, or Watchtower cycle (I hit exactly this bringing the container up: first response at ~105s, all of it the vite build).

The build genuinely has to run at start (not image-build time), because `vite build` inlines runtime-chosen client envs (`AUTH_MODE`, `DATABASE_PROVIDER`, `VITE_*`) into the client bundle, and the self-hoster only picks those at runtime via Compose. So the fix keeps the start-time build but **stops repeating it when nothing that affects the bundle changed**.

## Change

Move the start command into `docker-entrypoint.sh`, which:

1. Fingerprints the build-relevant env (`AUTH_MODE`, `DATABASE_PROVIDER`, all `VITE_*`) with `sha256sum`.
2. Reuses `dist/` when `dist/.openseo-build-env` matches that fingerprint (written only after a successful build).
3. Otherwise builds, then records the fingerprint.

`db:migrate:local` still runs every start (idempotent, cheap), unchanged.

## Behavior (verified)

| Scenario | Result |
| --- | --- |
| First start (no `dist/`) | **build** |
| Restart / host reboot, env unchanged | **skip → serve in seconds** |
| Change an inlined env (e.g. `AUTH_MODE`) | **build** (fingerprint changes) |
| Add/change a `VITE_*` var | **build** |
| Image update (fresh container, no `dist/`) | **build** |

The last row is the key safety property: an image update always lands a fresh container whose layer has no `dist/` (the image doesn't build at image-build time), so updated code always rebuilds — the cache never serves stale code across versions. Env changes are caught by the fingerprint even if the same container is restarted in place.

I unit-tested the decision block across all five scenarios above (fingerprint stability, change detection, skip vs rebuild) and confirmed `prettier --check .` still passes (it skips the shell/Dockerfile files).

## Notes / options

- Kept `pnpm run build` (which includes `tsc --noEmit`) as-is so the diff stays minimal and safe; the cache already skips it on restarts. If you'd rather also drop the runtime `tsc` (it re-type-checks the same shipped code every cold build), I'm happy to split the build script into `build` vs a runtime `build:app` in a follow-up.
- Happy to attach a before/after container timing (cold build vs cached restart) if useful.

### DIFF
--- a/Dockerfile.selfhost
+++ b/Dockerfile.selfhost
@@ -13,6 +13,8 @@ RUN pnpm install --frozen-lockfile
 
 COPY . .
 
+RUN chmod +x docker-entrypoint.sh
+
 EXPOSE 3001
 
 # Readiness probe against the unauthenticated setup/health endpoint. The long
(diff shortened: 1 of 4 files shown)


sorcerai:
Triage result: superseded by #120. This branch fingerprints the wrong build inputs (misses BYPASS_EMAIL_VERIFICATION, POSTHOG_PUBLIC_KEY, POSTHOG_HOST, TURNSTILE_SITE_KEY; includes non-inlined DATABASE_PROVIDER), has no persistent dist volume for container recreation, and has no tests. Maintainer action: close #123 and merge #120 after #153/check rerun.

bensenescu:
Great find @Santofer , thanks!

PR #137, opened 2026-07-23
fix(keywords): count rows actually written in refreshSavedKeywordMetrics

shuvamk:
## Problem

`refreshSavedKeywordMetrics` returns `{ updated }` to tell the caller how many saved keywords were refreshed. It computed:

```ts
updated += byKeyword.size;
```

`byKeyword` is keyed on the **provider's** returned metric keywords. But an upsert only happens when a *saved* keyword matches a returned metric:

```ts
const metric = byKeyword.get(r.row.keyword.toLowerCase());
if (!metric) return Promise.resolve(); // no write
```

So `updated` counts provider metrics, not rows written. The two diverge whenever the provider echoes a keyword string that doesn't equal a saved one — accent folding, casing differences, or metrics returned for keywords no longer saved — and can under-count when two returned metrics collapse to the same lower-cased key.

**Example:** one saved keyword `"café noir"`; the provider returns a metric for `"cafe noir"`. The match lookup misses, **0** rows are upserted, but the old code reports `{ updated: 1 }`.

The sibling `RankTrackingService.refreshKeywordMetrics` implements the same "refresh metrics, return `{updated}`" contract correctly by counting the actual writes (`updates.length`).

## Fix

Accumulate the number of upserts actually issued per chunk instead of the provider metric count.

## Tests

New `refresh-metrics.test.ts` (mocking the repository + dataforseo, matching the existing `saved-keywords.test.ts` pattern):
- provider returns 2 metrics but only 1 matches a saved keyword → `{ updated: 1 }` and exactly one upsert
- no saved keyword matches a returned metric → `{ updated: 0 }` and no upsert

Both fail on the old `byKeyword.size` logic (which returns 2 and 1).

## Verification

- `pnpm test` — 714 passed (90 files), including the new cases

### DIFF
--- a/src/server/features/keywords/services/research/refresh-metrics.test.ts
+++ b/src/server/features/keywords/services/research/refresh-metrics.test.ts
@@ -1,0 +1,88 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const mocks = vi.hoisted(() => ({
+  listSavedKeywordsByProject: vi.fn(),
+  upsertKeywordMetric: vi.fn(),
+  createDataforseoClient: vi.fn(),
+  fetchKeywordMetricsForList: vi.fn(),
+}));
+
+vi.mock(
+  "@/server/features/keywords/repositories/KeywordResearchRepository",
+  () => ({
+    KeywordResearchRepository: {
+      listSavedKeywordsByProject: mocks.listSavedKeywordsByProject,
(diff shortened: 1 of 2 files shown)


sorcerai (approved):
Reviewed exact current diff and tests. The result now counts actual writes rather than provider metrics, without changing the write set.

bensenescu:
Hey,

Thanks for contributing!

PR #138, opened 2026-07-23
fix(ai-search): drop null-metric rows when a brand-lookup range bound is set

shuvamk:
## Problem

`passesNumericFilter` in `brandLookupFiltering.ts` (used by the AI-search / brand-lookup citation tables) short-circuits on a null value **before** applying the bounds:

```ts
function passesNumericFilter(value, min, max) {
  if (value == null) return true; // returns before the min/max check
  ...
}
```

`mentions` and `aiSearchVolume` are nullable (`z.number().int().nonnegative().nullable()`). So a row whose metric is unknown passes through even when the user has set a "min mentions" / "min volume" bound — e.g. filtering queries with `minVolume: "100"` still shows rows with `aiSearchVolume: null`.

Every sibling numeric range filter in the app does the opposite once a bound is active:
- `RankTrackingFilters.logic.ts` `matchesMetricRangeFilter`: `if (!minValue && !maxValue) return true; if (value === null) return false;`
- the audit range filters exclude null rows too.

This function was the lone outlier, leaking rows the min/max UI implies should be hidden.

## Fix

Return early only when **no** bound is set; once a bound is active, exclude rows whose value is null — matching the established `matchesMetricRangeFilter` convention.

## Tests

`brandLookupFiltering` had no tests. Added `brandLookupFiltering.test.ts`:
- `filterQueries` with `minVolume: "100"` drops a `aiSearchVolume: null` row (fails on the old early-return)
- with no numeric bound, the null row is kept
- `filterTopPages` with `minMentions: "5"` drops a `mentions: null` row

## Verification

- `pnpm test` — 713 passed (88 files), including the new cases

### DIFF
--- a/src/client/features/ai-search/brandLookupFiltering.test.ts
+++ b/src/client/features/ai-search/brandLookupFiltering.test.ts
@@ -1,0 +1,63 @@
+import { describe, expect, it } from "vitest";
+import { filterQueries, filterTopPages } from "./brandLookupFiltering";
+import {
+  EMPTY_QUERIES_FILTERS,
+  EMPTY_TOP_PAGES_FILTERS,
+} from "./brandLookupFilterTypes";
+import type { BrandLookupResult } from "@/types/schemas/ai-search";
+
+type TopQuery = BrandLookupResult["topQueries"][number];
+type TopPage = BrandLookupResult["topPages"][number];
+
+function topQuery(question: string, aiSearchVolume: number | null): TopQuery {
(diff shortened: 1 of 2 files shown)


sorcerai (approved):
Reviewed exact current diff and tests. Null metrics are excluded only when a numeric bound is active, matching existing filter semantics.

bensenescu:
Hey,

Thanks for contributing!

PR #139, opened 2026-07-23
fix(mcp): send local-business search radius in meters, not kilometers

shuvamk:
## Problem

The `search_local_businesses` MCP tool built its DataForSEO `business_data` `location_coordinate` like this:

```ts
function formatBusinessLocationCoordinate(near) {
  return `${formatCoordinate(near.latitude)},${formatCoordinate(near.longitude)},${near.radiusKm}`;
}
```

`radiusKm` is documented (and validated) as kilometers, but DataForSEO's `business_data` `location_coordinate` expects the radius in **meters** (200–199999). So the value was passed through 1000× too small: a `radiusKm: 5` request became a **5-meter** radius and returned almost nothing.

The sibling formatter for the *same* `nearSchema` — Google Business Q&A — already did the right thing:

```ts
const radius = Math.min(MAX, Math.max(MIN, Math.round(near.radiusKm * 1000))); // km -> m
```

The two contradicted each other; meters is correct (the 200–199999 clamp range is the `business_data` meter constraint).

## Fix

Both formatters are for `business_data` endpoints and, once corrected, are identical — so unify them into one `formatBusinessDataCoordinate` that converts kilometers to meters and clamps to the provider's range. Both call sites (`search_local_businesses`, `get_google_business_questions`) use it.

## Tests

The existing `dataforseo-research-tools.test.ts` asserted the un-converted value (`,5`) for the local-business path, locking in the bug — updated to `,5000`, matching the Q&A test that already asserts `,5000` for the same `radiusKm: 5`.

## Verification

- `pnpm test` — 710 passed (87 files)

### DIFF
--- a/src/server/mcp/tools/dataforseo-research-tools.test.ts
+++ b/src/server/mcp/tools/dataforseo-research-tools.test.ts
@@ -110,7 +110,7 @@ describe("DataForSEO research MCP tools", () => {
 
     expect(businessListings).toHaveBeenCalledWith(
       expect.objectContaining({
-        locationCoordinate: "33.1234568,-84.9876543,5",
+        locationCoordinate: "33.1234568,-84.9876543,5000",
         categories: ["cafe"],
       }),
     );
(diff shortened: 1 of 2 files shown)


sorcerai (approved):
Reviewed exact current diff and provider-payload test. The shared formatter sends the business-data radius in meters and preserves the documented clamp.

bensenescu:
Hey,

Thanks for contributing!

PR #141, opened 2026-07-23
fix(dashboard): count rank summary per keyword, not per device

shuvamk:
## Problem

`getRankSummary` builds the dashboard rank scorecard. It counts the denominator per keyword but the numerators per device:

```ts
summary.trackedKeywords += result.rows.length;        // one row = one keyword
for (const row of result.rows) {
  for (const device of ["desktop", "mobile"] as const) {  // both devices
    const { position, previousPosition } = row[device];
    if (position !== null && position <= 10) summary.top10 += 1;
    ...
  }
}
```

Each row carries both a `desktop` and a `mobile` result for the **same** keyword, so a keyword ranking in the top 10 on both devices adds **2** to `top10` but **1** to `trackedKeywords`. The summary can therefore report *more top-10 keywords than tracked keywords* (e.g. "2 in top 10 out of 1 tracked"). `improved`/`declined` have the same cross-basis inflation.

The tested sibling `computeScorecards` computes these per keyword **for a single device** (it takes a `device` param), confirming the intended unit is per-keyword — not both devices summed into a per-keyword denominator.

## Fix

Extract a pure `summarizeRankRows` helper that classifies each keyword **once**, using its best (lowest) position across devices, and use it in `getRankSummary`. All four numbers now share the per-keyword basis, so `top10`/`improved`/`declined` can never exceed `trackedKeywords`.

**Semantic note for reviewers:** I collapsed the two devices to the keyword's best position — the natural reading for an overview scorecard ("how many of my keywords rank in the top 10"). The existing null-handling is preserved (a keyword only counts toward improved/declined when it has both a current and a previous position). If you'd prefer a different aggregation (e.g. a fixed primary device, or counting "new"/"lost" like `computeScorecards` does), the helper is the single place to change it.

## Tests

New `rankSummary.test.ts`: a both-device top-10 keyword counted once; `top10` never exceeding `trackedKeywords`; best-device position used for the top-10 check; a null-current keyword ignored; empty input zeroed.

## Verification

- `pnpm test` — 715 passed (88 files)

### DIFF
--- a/src/server/features/dashboard/services/DashboardService.ts
+++ b/src/server/features/dashboard/services/DashboardService.ts
@@ -3,6 +3,7 @@ import { ActivationRepository } from "@/server/features/activation/repositories/
 import { AuditRepository } from "@/server/features/audit/repositories/AuditRepository";
 import { getIssueTypePageCountsForAudit } from "@/server/features/audit/repositories/auditSummaryQueries";
 import { BacklinkSnapshotRepository } from "@/server/features/dashboard/repositories/BacklinkSnapshotRepository";
+import { summarizeRankRows } from "@/server/features/dashboard/services/rankSummary";
 import { GscConnectionRepository } from "@/server/features/gsc/repositories/GscConnectionRepository";
 import { RankTrackingRepository } from "@/server/features/rank-tracking/repositories/RankTrackingRepository";
 import { getLatestResults } from "@/server/features/rank-tracking/services/rankTrackingResults";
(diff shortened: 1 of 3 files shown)


sorcerai (approved):
Reviewed exact current diff and tests. Summary counts each keyword once using its best device position, restoring top10 <= trackedKeywords.

bensenescu:
Hey,

Thanks for contributing!

Issue #50, opened 2026-06-30
Rank Tracking Page - Add paginated domain keyword discovery 


bensenescu:
## Problem
Rank Tracking should let users fetch keywords their domain already ranks for from the rank tracking page, paginate through those results, and choose which ones to track. The current rank tracking flow only exposes a manual add panel on the domain detail page, while the existing ranked-keyword suggestion flow appears limited to setup and top results.

## Evidence
- User feedback: "On the rank tracking page, there should be an option to fetch keywords for your domain, this will get a list of the top keywords your domain currently ranks for which the user can paginate through. It should be similar to the domain overview page and the current implemetation in rank tracking may be slightly wrong."

## Implementation check
Checked the OpenSEO repo on `main` (`0e33e8aa`). Relevant files:
- `src/client/features/rank-tracking/RankTrackingDomainDetail.tsx`: domain detail page has an “Add keywords” toggle and toolbar actions for checking rankings / refreshing metrics, but no paginated domain-keyword discovery action.
- `src/client/features/rank-tracking/AddKeywordsPanel.tsx`: current detail-page add flow is manual textarea entry only.
- `src/client/features/rank-tracking/KeywordSuggestionStep.tsx`: setup flow fetches ranked-keyword suggestions, but it uses `getDomainKeywordSuggestions`, preselects the top 20, and does not expose paginated browsing.
- `src/serverFunctions/domain.ts` and `src/server/features/domain/services/domainKeywordsPage.ts`: Domain Overview already has a paginated ranked-keywords endpoint with sorting/filtering inputs.
- `src/server/features/domain/services/DomainService.ts`: `getSuggestedKeywords` is a separate top-100 helper, which is likely the wrong abstraction for the paginated Rank Tracking page flow.

This is ready for implementation because the paginated domain-keyword data path and the rank-tracking keyword add path already exist; the work is to compose them into the rank tracking UI and retire/avoid the non-paginated suggestion flow where it is the wrong fit.

## Suggested direction
Add a “Find domain keywords” / “Fetch domain keywords” action from the rank tracking domain detail page. Reuse the Domain Overview ranked-keywords pagination path for the config’s domain/location/language, let users select keywords across pages, and add selected terms through the existing `addTrackingKeywords` flow. The UI should make already-tracked keywords clear and avoid adding duplicates.

## Acceptance criteria
- [ ] Rank Tracking domain detail page has an option to fetch/discover keywords the configured domain currently ranks for.
[...]

PR #143, opened 2026-07-23
fix(rank-tracking): add Check Now button to empty state

Alexi5000:
Related to #50

## Problem
When a rank tracker has no keywords yet, the table empty state says to click "Check Now", but that action only lives in the overflow menu next to Export — so the empty state points at a control that isn't visible.

## Fix
- Add a primary Check Now button on that empty state (same handler as the More menu).
- If there are still no keywords, open Add keywords and toast instead of starting an empty check.

## Verification
- `pnpm exec prettier --write` on the two touched files
- `pnpm exec tsc --noEmit`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/client/features/rank-tracking/RankTrackingDomainDetail.tsx
+++ b/src/client/features/rank-tracking/RankTrackingDomainDetail.tsx
@@ -186,6 +186,16 @@ export function RankTrackingDomainDetail({
   // Fall back to the table if history disappears (e.g. device switch).
   const effectiveViewMode = historyAvailable ? viewMode : "table";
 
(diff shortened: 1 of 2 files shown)


sorcerai (changes requested):
The empty-state copy still says Click Check Now for free-plan users, but the button is hidden when checkDisabled/isFreePlan is true. That leaves the original broken instruction for that segment. Please make the copy conditional or render a disabled upgrade affordance, add paid/free render coverage, and rerun the currently failing checks.

bensenescu:
Hey,

Thanks for contributing!

PR #151, opened 2026-07-25
fix(audit): stop applying the cumulative capacity cap to self-hosted deploys

Nordalux:
## Problem

`resolveAuditLimitTier` resolves self-hosted deploys to the `paid` tier:

```ts
// src/server/features/audit/services/AuditService.ts
if (!(await isHostedServerAuthMode())) return "paid";
```

`paid` carries `maxCapacityUnits: 100_000`, and that bound is **cumulative, not concurrent**. `getAuditUsageForUser` sums over every audit row the user has ever started, with no status filter and no time window:

```ts
// src/server/features/audit/repositories/AuditRepository.ts:358
const rows = await db.query.audits.findMany({
  where: eq(audits.startedByUserId, userId),
  columns: { status: true, pagesTotal: true, lighthouseTotal: true },
});

return {
  capacityUnits: rows.reduce(
    (total, row) => total + row.pagesTotal + row.lighthouseTotal,
    0,
  ),
  runningCount: rows.filter((row) => row.status === "running").length,
};
```

On hosted that is exactly the intended abuse control, and I'm not proposing to change it there. On a self-hosted deploy it becomes a lifetime budget on the operator's own hardware:

| audit size | audits before `AUDIT_CAPACITY_REACHED` |
| ---------- | -------------------------------------- |
| 10,000 pages | ~10 |
| 500 pages | ~190 |
| 50 pages | ~2,000 |

Units are reserved from the **planned** page count, not the crawled one, so an audit configured for 10,000 pages that finishes after 300 still spends 10,020 units. Once the budget is gone, the only way to reclaim it is deleting old audit rows.

The self-hosted crawler runs on the operator's own compute and isn't credit-metered, so this bound isn't protecting anything.

## Change

Add a `self_hosted` tier with `maxCapacityUnits` lifted, and resolve non-hosted deploys to it. This matches how every other hosted-only gate already treats self-hosted — credits (`SamChatAgent`, `OnboardingChatAgent`), plan checks (`HostedPlanGate`), and DataForSEO metering (`meterDataforseoCall`) all skip it, with the existing code comments stating the intent explicitly:

> Self-hosted brings its own provider keys and has no Autumn balance, so it's ungated.

`maxPagesPerAudit` deliberately stays at the paid value. That one is a real technical ceiling (Workflow step budget, D1 row writes), not a plan limit, and a test pins it so it doesn't drift.

## Hosted behaviour

Unchanged. `free` and `paid` keep their existing bounds, and a hosted deploy can never resolve to the new tier. The one place the tier value leaves the server is the `plan_tier` property on the `site_audit:start` PostHog event — `captureServerEvent` no-ops when not hosted, so the new value can't reach analytics.

## Tests

- `audit-capacity.test.ts` — asserts the hosted tiers keep a finite cumulative bound while `self_hosted` does not, and that `self_hosted.maxPagesPerAudit` stays identical to `paid`.
- `AuditService.limitTier.test.ts` (new) — covers all four resolver paths: self-hosted resolves to `self_hosted` without consulting billing at all, hosted subscriber to `paid`, hosted without a paid plan to `free`, and hosted without managed access rejects with `PAYMENT_REQUIRED`.

`pnpm ci:check` passes. Full `vitest run` is green apart from `src/server/lib/dataforseo/client.test.ts > skips billing in non-hosted mode`, which also fails on an unmodified `main` on my machine (a 5s timeout under full-suite load, ~4.2s in isolation) and is unrelated to this change.

## Not in scope

Whether the cumulative counting in `getAuditUsageForUser` is the right shape for the hosted free tier is a separate question — a rolling window or an active-only filter would both be defensible. That's a behaviour change for paying-adjacent users and belongs in its own issue rather than riding along here.

### DIFF
--- a/src/server/features/audit/services/AuditService.limitTier.test.ts
+++ b/src/server/features/audit/services/AuditService.limitTier.test.ts
@@ -1,0 +1,77 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const { isHostedMock, hasManagedAccessMock, hasPaidPlanMock } = vi.hoisted(
+  () => ({
+    isHostedMock: vi.fn(),
+    hasManagedAccessMock: vi.fn(),
+    hasPaidPlanMock: vi.fn(),
+  }),
+);
+
+vi.mock("cloudflare:workers", () => ({ env: {} }));
+
+vi.mock("@/server/lib/runtime-env", () => ({
+  isHostedServerAuthMode: isHostedMock,
+}));
+
+vi.mock("@/server/billing/subscription", () => ({
+  customerHasManagedAccess: hasManagedAccessMock,
+  customerHasPaidPlan: hasPaidPlanMock,
+}));
+
+// Pulled in transitively by AuditService; stubbed so the tier resolver can be
+// exercised without a database or a KV binding.
+vi.mock("@/server/features/audit/repositories/AuditRepository", () => ({
+  AuditRepository: {},
+}));
+
+vi.mock("@/server/lib/audit/progress-kv", () => ({
+  AuditProgressKV: {},
+}));
+
+import { AuditService } from "@/server/features/audit/services/AuditService";
+import { AppError } from "@/server/lib/errors";
+
+describe("resolveAuditLimitTier", () => {
+  beforeEach(() => {
+    hasManagedAccessMock.mockResolvedValue(true);
+    hasPaidPlanMock.mockResolvedValue(true);
+  });
+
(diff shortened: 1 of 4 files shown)


sorcerai (approved):
Reviewed exact current diff. Self-hosted installs no longer inherit the hosted cumulative commercial quota; the per-audit technical ceiling remains and hosted behavior is unchanged.

bensenescu:
Hey,

Thanks for contributing!

PR #52, opened 2026-07-01
feat: free Lighthouse audits via PageSpeed Insights when PAGESPEED_API_KEY is set

mattmacrocket:
## What

Site-audit Lighthouse currently runs on the metered DataForSEO OnPage endpoint ($0.00425/page) even though Google's PageSpeed Insights API returns the same Lighthouse report for free (roughly 25k requests/day with a free API key). This PR routes the audit's Lighthouse phase to PSI **when a `PAGESPEED_API_KEY` is set**, and falls back to the existing DataForSEO path otherwise, so current deployments keep their exact behavior with zero action.

- New `src/server/lib/pagespeedLighthousePayload.ts`: fetches PSI and normalizes the report into the exact same `StoredLighthousePayload` the DataForSEO path produces. R2 payloads, scores, metrics, and the issue-details view are provider-agnostic; the stored-payload `source` enum accepts both providers and existing stored payloads still parse.
  - Handles a Lighthouse 13.x quirk: `details.items` comes back as an object (not array) for some audits (e.g. `document-latency-insight`), which would otherwise fail the whole parse.
  - 90s timeout (PSI runs can be slow), same 3-attempt retry loop as before.
- Plumbs `PAGESPEED_API_KEY` through `.env.example`, `compose.yaml`, `env.d.ts`, and documents it in `docs/SELF_HOSTING_DOCKER.md`.

## Why opt-in via key rather than always PSI

Keyless PSI requests share a tiny anonymous global quota that is almost always exhausted (HTTP 429 in practice), so PSI is only reliable with a key. Gating on the key means: self-hosters get free audits with a 2-minute free key, and deployments that do nothing (including hosted) are completely unaffected.

## Notes for review

- **Hosted:** setting `PAGESPEED_API_KEY` on the hosted deployment would make audit Lighthouse free there too (no DataForSEO OnPage spend, no user credit metering on that path). Without the key, hosted behavior is unchanged.
- The old DataForSEO Lighthouse client path stays intact (it is the fallback). If you ever want PSI-only, `client.lighthouse.live` and `dataforseoLighthousePayload.ts` could be removed in a follow-up.

## Verification

- Live end-to-end audit (10 pages, Lighthouse on, key set): 8/8 Lighthouse runs via PSI (4 sampled pages x mobile/desktop), scores and per-page issue details render unchanged, zero DataForSEO credits spent (account balance verified before/after).
- Fallback path exercised by unsetting the key (identical to current behavior).
[...]

PR #152, opened 2026-07-25
fix(docker): forward OPENROUTER_API_KEY into the self-host container

Nordalux:
## Problem

`OPENROUTER_API_KEY` is documented as the key that unlocks the AI features on a self-hosted install:

- `.env.example`: _"Optional in self-hosted modes. Required if you want AI features like SAM, the in-app SEO agent."_
- `src/routes/_app/help/openrouter-api-key.tsx` — a whole in-app help page for obtaining and setting it.
- `src/client/features/sam/SamSetupGate.tsx` — tells the user to set the `OPENROUTER_API_KEY` environment variable and restart.

Under Docker none of that has any effect. Compose only forwards the keys listed under `environment:` in `compose.yaml`, and this one was not listed. `SamChatAgent.ts:110` calls `getEnvValueSync(this.env, "OPENROUTER_API_KEY")`, gets nothing, and throws `OPENROUTER_API_KEY is required for the SAM agent` — so the user follows the in-app instructions, restarts, and SAM stays broken with no indication why.

## Change

Add the passthrough next to the existing optional Google Search Console keys, using the same `${VAR:-}` form so it stays optional.

## Verification

With a `.env` containing `OPENROUTER_API_KEY=sk-or-test123`:

```
# main
$ docker compose -f compose.yaml config | grep OPENROUTER
(no output)

# this branch
$ docker compose config | grep OPENROUTER
      OPENROUTER_API_KEY: sk-or-test123
```

`pnpm ci:check` passes.

## Docs

`SELF_HOSTING_DOCKER.md` said Compose "passes `.env` values into the container", which reads as though any key in `.env` arrives in the container — the assumption that makes this bug easy to miss. Corrected to say it forwards the keys `compose.yaml` lists, and added `OPENROUTER_API_KEY` to the optional env values list.

## Note

#115 adds `CHAT_BASE_URL` for local/OpenAI-compatible chat models. If that lands, it will need the same passthrough line to work under Docker — happy to add it here or leave it to that PR, whichever the maintainers prefer.


sorcerai (approved):
Reviewed exact current diff. OPENROUTER_API_KEY is forwarded without logging or changing defaults, and the Docker documentation now matches Compose behavior.

sorcerai:
Integration note: #52 and this PR both touch compose.yaml and docs/SELF_HOSTING_DOCKER.md; merging either second conflicts. Safest order is #52 first, then rebase this small PR. While rebasing, please take up your offer from the PR body and add CHAT_BASE_URL passthrough for #115 so local/OpenAI-compatible chat also works under Docker.

PR #159, opened 2026-07-28
feat(ai): support any OpenAI-compatible LLM gateway alongside OpenRouter

knowhycodata:
## What

SAM and the onboarding agent could only talk to OpenRouter. This adds a second, opt-in path: set `LLM_API_KEY` and OpenSEO talks to any endpoint that speaks the OpenAI chat-completions API.

- `LLM_BASE_URL` defaults to `https://llmtr.com/v1` ([LLMTR](https://llmtr.com), a hosted gateway that serves the same `vendor/model` slugs). Point it at another gateway, or at your own vLLM/Ollama server.
- `LLM_MODEL` defaults to `minimax/minimax-m3`, the same slug the OpenRouter path already defaults to. It resolves on both.
- `OPENROUTER_API_KEY` behaves exactly as before and remains the default path. Nothing changes for existing deployments.
- If both keys are set, `LLM_API_KEY` wins — so a deployment can switch over without first deleting the old secret.

## Why

Self-hosters who would rather not route through OpenRouter had no way in short of patching the provider module. That is a real constraint for anyone with an existing gateway contract, a regional/data-residency requirement, or a local model they already run.

## How

Provider selection moves into `src/server/lib/chatAgentModel.ts` (replacing `openrouter.ts`), shared by both chat agents and the SAM setup gate. `resolveChatAgentProvider(env)` returns a discriminated union; `buildChatAgentModel` maps it to a `LanguageModelV3` via `@openrouter/ai-sdk-provider` or `@ai-sdk/openai-compatible`.

The OpenRouter request config is untouched and stays on that path only — usage accounting, the ZDR provider ordering, and the reasoning channel are all OpenRouter-specific and not portable to a plain chat-completions endpoint.

## Trade-off worth flagging

Cost metering reads `providerMetadata.openrouter.usage.cost`, which only OpenRouter reports. Deployments on the gateway path therefore meter **no** LLM spend against the credit pool and track cost in their gateway's own dashboard instead. That is stated in both the in-app help page and the self-hosting docs. It suits the self-hosted case this path is for; hosted deployments stay on OpenRouter.

`@ai-sdk/openai-compatible` is pinned to the `2.x` line, which builds on `@ai-sdk/provider@3` — the `3.x` line moved to provider v4 / `LanguageModelV4` and does not match the `ai@6` / Think stack here.

## Test plan

- New unit tests in `src/server/lib/chatAgentModel.test.ts` cover provider selection: neither key set, each key alone, both set, and the base-URL/model overrides.
- `pnpm ci:check` — `knip`, `tsc --noEmit`, and `oxlint --type-aware` all clean.
- `vitest run` — 740 tests across 92 files pass.
- `vite build` succeeds.

Note: `prettier --check .` fails on this Windows checkout for ~790 files it also fails on before this branch (CRLF vs LF, no `.gitattributes` in the repo). Logged in `.agents/PAPERCUTS.md` rather than fixed here, since a line-ending sweep does not belong in this PR. Touched files were formatted with prettier individually.


bensenescu:
Hey,

Thanks for contributing!

knowhycodata:
Rebased onto current `main` (`fb4b76e`) — the PR was `CONFLICTING` and is now mergeable.

Two conflicts, both resolved without dropping anyone's work:

- **`.agents/PAPERCUTS.md`** — an append-only log. `main` had added two newer entries while this branch added one. Kept all three, in the file's existing reverse-chronological order.
- **`pnpm-lock.yaml`** — not hand-merged. I took `main`'s lockfile wholesale and regenerated with `pnpm install --lockfile-only`, so the diff is only what this PR actually adds: `@ai-sdk/openai-compatible` and its transitive deps, +37 lines, additive. (Per the `minimumReleaseAge` papercut already in PAPERCUTS.md, regeneration needs `--config.minimumReleaseAge=0` or the gate re-blocks already-pinned peers.)

Verified after the rebase, not just assumed:

```
pnpm run types:check   -> clean
pnpm run test          -> 124 files, 1013 tests, all passing
```

The `UNSTABLE` status is from the same checks that were already amber before the rebase, not from anything this update introduced. Happy to rebase again whenever it drifts.

PR #164, opened 2026-07-28
refactor: extract the shared GSC MCP helpers into gsc-shared.ts

mattbalza:
`search-console-tools.ts` opens with `/* eslint-disable max-lines */`. Three of
its helpers are about Search Console *connection state* rather than about either
tool defined in the file:

- `connectGscUrl` — where to send someone who needs to connect
- `missingSelfHostedGoogleClientResponse` — the self-hosted setup nudge
- `describeGscError` — the `GscNotConnectedError` / `GscTokenError` / `GscApiError` wording

Moving them into `gsc-shared.ts` is what that disable comment was standing in
for, and the file drops back under the cap so the suppression goes with them.

## No behaviour change

The three functions move verbatim. The file-local `ProjectAuthContext` type moves
with them as `GscProjectAuthContext`. Nothing else in the file changes beyond the
import block.

The payoff is for the next GSC-backed MCP tool: today it would have to restate
"not connected", "token expired", and the setup nudge, and the three copies would
drift. Downstream I have four such tools sharing this module, which is how the
extraction came up.

## Test plan

```bash
pnpm ci:check    # prettier + knip + tsc + oxlint --type-aware
pnpm test
```

`knip` is the meaningful one here — it would flag any of the three exports if the
move left a caller behind.

### DIFF
--- a/src/server/mcp/tools/gsc-shared.ts
+++ b/src/server/mcp/tools/gsc-shared.ts
@@ -1,0 +1,58 @@
+import { buildProjectMeta } from "@/server/mcp/context";
+import { mcpResponse } from "@/server/mcp/formatters";
+import { buildDashboardUrl } from "@/server/mcp/urls";
+import { hasSelfHostedGscConfig } from "@/server/features/gsc/oauth-config";
+import { isHostedServerAuthMode } from "@/server/lib/runtime-env";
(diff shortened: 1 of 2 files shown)


bensenescu:
Hey,

Thanks for contributing!

PR #166, opened 2026-07-29
fix(audit): keep blank rows last when sorting results tables descending

evanclan:
## Summary

Sorting any nullable column in the Site Audit **Pages** or **Performance** table descending moves the blank rows to the *top* instead of leaving them at the bottom. The two comparators those tables use return `±1` for a blank side, and TanStack negates a comparator's result on a descending column, so the intent is inverted on the second header click.

This reuses the helper the repo already has for exactly this problem — `src/client/components/table/nullSafeSort.ts` — instead of keeping a second, direction-blind copy. 11 sortable columns are affected.

## Problem

`getSortedRowModel` applies the descending flip *after* calling the column's `sortingFn` ([`table-core@8.21.3`](https://github.com/TanStack/table/blob/main/packages/table-core/src/utils/getSortedRowModel.ts)):

```ts
if (sortInt === 0) {
  sortInt = columnInfo.sortingFn(rowA, rowB, sortEntry.id)
}

// If sorting is non-zero, take care of desc and inversion
if (sortInt !== 0) {
  if (isDesc) {
    sortInt *= -1
  }
```

So a comparator that hard-codes "blank goes after" gets "blank goes before" on desc:

```ts
// src/client/features/audit/results/AuditResultsTableFilterLogic.ts:167
export function nullableNumberSort(left, right, columnId) {
  const a = left.getValue(columnId);
  const b = right.getValue(columnId);
  if (a == null && b == null) return 0;
  if (a == null) return 1;   // becomes -1 when the column is desc
  if (b == null) return -1;
  return a - b;
}
```

Verified against a real table instance — with rows `score: [5, null, 10]` and the column sorted desc, the old comparator returns `[null, 10, 5]`.

The user-visible sting is on the Performance tab. `isLighthouseFailure` is defined as "has no category scores" (`AuditResultsTableFilterLogic.ts:60-71`), so the null rows *are* the failed audits. Clicking **Perf**, **A11y** or **SEO** to see your worst pages first fills the top of the table with rows that have no scores at all, pushing the real data below the fold. Same for LCP / CLS / INP / TTFB, for **Status** and **Title** on the Pages tab (pages that aren't HTML documents have no title), and for the Performance **URL** column.

Note that `sortUndefined: "last"` is not an option here: it returns early *before* the flip, so it is direction-safe, but it only inspects `undefined` and these columns carry `null` straight from the DB.

## Solution

`nullSafeSort.ts` already solves this for the Brand Lookup tables (added in `faf7a29`) by reading the column's sort direction off the cell and pre-compensating for the flip. This PR:

- adds `stringNullsLast`, a sibling of the existing `numericNullsLast`, for the URL/Title columns — empty strings count as blank, preserving the old `!a` behaviour;
- factors the shared "one side is blank" sign into a small `blankLast` helper used by both;
- points the 3 Pages columns and 8 Performance columns at the shared helpers;
- deletes `nullableNumberSort` / `nullableStringSort`, so there is one implementation of this rule rather than two.

There is a second working pattern in the repo — `RankTrackingColumns.tsx` maps `accessorFn: (row) => row.cpc ?? undefined` and lets `sortUndefined: "last"` do the work. That would also fix these tables, but it needs an `accessorFn` on all 11 columns, so reusing the existing comparator seemed like the smaller change. Happy to switch if you'd rather converge on the rank-tracking approach.

## Out of scope

- `KeywordSuggestionStep.tsx:70-124` has the same class of issue via sentinels (`position ?? 999`, `searchVolume ?? 0`), so unranked suggestions lead on a descending position sort. Different feature and it needs a call on whether `0` should mean "unknown" for volume/traffic, so I left it alone.
- No change to the audit filter logic, the tables' markup, or any other feature's sorting.

## Test plan

- [x] New `src/client/components/table/nullSafeSort.test.ts` drives a real headless `createTable` + `getSortedRowModel`, so it exercises TanStack's own desc flip rather than asserting my reading of it. Covers nulls-last in both directions, `0` not treated as blank, blanks keeping their relative order, and `""` counting as blank.
- [x] Confirmed the new tests fail against the removed comparators (blank sorts first on desc) and pass with the shared ones.
- [x] `pnpm test` — 92 files, 740 tests passing.
- [x] `pnpm ci:check` — prettier, knip, `tsc --noEmit` (both projects) and `oxlint --type-aware` all clean.

## Related

- `src/client/components/table/nullSafeSort.ts` — the existing fix for this bug class, used by the Brand Lookup citation tables.
- `src/client/features/rank-tracking/RankTrackingColumns.tsx` — the `null` -> `undefined` + `sortUndefined: "last"` variant of the same fix.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/client/components/table/nullSafeSort.test.ts
+++ b/src/client/components/table/nullSafeSort.test.ts
@@ -1,0 +1,105 @@
+import {
+  createTable,
+  getCoreRowModel,
+  getSortedRowModel,
+  type ColumnDef,
+} from "@tanstack/react-table";
+import { describe, expect, it } from "vitest";
+import {
+  numericNullsLast,
+  stringNullsLast,
+} from "@/client/components/table/nullSafeSort";
+
+type TestRow = { id: string; score: number | null; title: string | null };
+
+const columns: ColumnDef<TestRow>[] = [
+  { id: "score", accessorKey: "score", sortingFn: numericNullsLast },
+  { id: "title", accessorKey: "title", sortingFn: stringNullsLast },
+];
+
+/**
+ * Sort through a real table instance rather than calling the comparators
+ * directly: TanStack negates a comparator's result on a descending column, and
+ * surviving that flip is the whole point of these helpers.
+ */
+function sortedIds(
+  rows: TestRow[],
+  columnId: "score" | "title",
+  desc: boolean,
+): string[] {
+  const table = createTable<TestRow>({
+    data: rows,
+    columns,
+    state: { sorting: [{ id: columnId, desc }] },
+    onStateChange: () => {},
+    renderFallbackValue: null,
+    getCoreRowModel: getCoreRowModel(),
+    getSortedRowModel: getSortedRowModel(),
+  });
+  return table.getSortedRowModel().rows.map((row) => row.original.id);
+}
+
+describe("numericNullsLast", () => {
+  const rows: TestRow[] = [
+    { id: "mid", score: 5, title: "beta" },
+    { id: "blank", score: null, title: null },
+    { id: "high", score: 10, title: "alpha" },
+  ];
+
+  it("orders ascending with nulls last", () => {
+    expect(sortedIds(rows, "score", false)).toEqual(["mid", "high", "blank"]);
+  });
+
+  it("keeps nulls last when descending", () => {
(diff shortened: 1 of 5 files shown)


bensenescu:
Hey,

Thanks for contributing!

PR #170, opened 2026-07-29
fix(audit): stop re-billing Lighthouse on workflow and parse retries

Ayush7614:
## Summary
- Site-audit Lighthouse batches used default Cloudflare Workflow retries (limit 5). A mid-batch failure after DataForSEO already charged (e.g. R2/D1 write) re-ran the whole batch and billed again.
- Post-`assertOk` Lighthouse parse failures threw plain `Error`, so Autumn metering skipped the spend and in-process retries called DataForSEO again.
- Aligns metered Lighthouse steps with rank-check’s single-attempt pattern, meters charged parse failures, and soft-fails instead of retrying billed errors.

## How to review
1. `siteAuditWorkflowPhases.ts` — `LIGHTHOUSE_BATCH_STEP_CONFIG` with `retries.limit: 0`
2. `dataforseo/lighthouse.ts` — billing built before parse; parse failures become `DataforseoChargedTaskError`
3. `audit/lighthouse.ts` — break the 3-attempt loop on charged errors
4. Unit test covers the parse-after-bill path

## Test plan
- [x] `pnpm exec vitest run src/server/lib/dataforseo/lighthouse.test.ts src/server/lib/audit/lighthouse.test.ts`
- [x] Spot-check a hosted site audit Lighthouse batch failure does not re-fetch already-charged URLs

### DIFF
--- a/src/server/lib/audit/lighthouse.ts
+++ b/src/server/lib/audit/lighthouse.ts
@@ -1,6 +1,7 @@
 import { detectUrlTemplate, canonicalUrlKey } from "./url-utils";
 import type { BillingCustomerContext } from "@/server/billing/subscription";
 import { createDataforseoClient } from "@/server/lib/dataforseo";
+import { DataforseoChargedTaskError } from "@/server/lib/dataforseo/envelope";
 import type { LighthouseResult, LighthouseStrategy } from "./types";
 import { putTextToR2 } from "@/server/lib/r2";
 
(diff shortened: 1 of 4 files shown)


bensenescu:
thanks for raising this. ended up fixing it in this commit: https://github.com/every-app/open-seo/commit/211907ae32e5ebaefeba169bcae1e8baeacb1289

PR #171, opened 2026-07-29
fix(rank-tracking): advance schedule when skipping unpaid cron configs

Ayush7614:
## Summary
- Hosted scheduled rank-check cron loads up to 50 due configs, then skipped unpaid orgs with `continue` and never advanced `nextCheckAt`.
- Those rows stayed due forever, filled the limit, and could starve paying orgs of scheduled checks.
- Unpaid skips now advance the schedule the same way empty-keyword skips already do.

## How to review
1. `scheduledRankChecks.ts` — unpaid path calls shared `advanceNextCheckAt`
2. Unit test: unpaid config is rescheduled; paid config still starts a run

## Test plan
- [x] `pnpm exec vitest run src/server/features/rank-tracking/services/scheduledRankChecks.test.ts`
- [x] Confirm a hosted unpaid due config is pushed forward on the next cron tick without starting a workflow

### DIFF
--- a/src/server/features/rank-tracking/services/scheduledRankChecks.test.ts
+++ b/src/server/features/rank-tracking/services/scheduledRankChecks.test.ts
@@ -1,0 +1,104 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const mocks = vi.hoisted(() => ({
+  getDueConfigsWithOrganization: vi.fn(),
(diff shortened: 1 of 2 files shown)


bensenescu:
Hey Ayush, thanks for raising this. I ended fixing this in this commit: https://github.com/every-app/open-seo/commit/f048dc3bd1c671e28aa4ceaaf22ca6663cdcb321

PR #180, opened 2026-07-30
feat(rank-tracking): opt-in env toggles to run manual rank checks via the task queue

redrammedia:
## Motivation

Scheduled rank checks already use DataForSEO's standard task queue (~30% of the live endpoint's cost), but manual "Check now" runs always pay live prices. For self-hosters who track a few hundred keywords and don't mind waiting a few minutes, that's a ~3.3x cost difference with identical data — both paths send the same request parameters and parse results through the same `buildRankCheckResult`.

This PR adds two opt-in env toggles so self-hosters can choose cost over speed. **Default behavior is completely unchanged** — without the variables set, every code path resolves to today's behavior.

## Changes

**`RANK_CHECK_MANUAL_METHOD=queued`** routes manual checks through the existing `runQueuedCheck` path (task_post → task_get polling → live fallback). The decision is resolved inside the durable "prepare" workflow step via a new `resolveRankCheckMethod()` helper, so it survives workflow replays. The credit pre-check estimates at queued pricing accordingly.

**`RANK_CHECK_QUEUE_LIVE_FALLBACK=off`** disables the live fallback for queued stragglers (some self-hosters prefer "never bill live prices" over completeness). To compensate, the polling window extends from ~15 minutes to ~2¼ hours (cumulative waits 4/6/8/10/12/15/30/45/60/75/105/135 min). Anything still missing afterwards is reported by the existing finalize logic as "could not be checked" instead of being re-billed live.

**Cost estimate follows the resolved method:** `RankTrackingService.estimateCost` and the check-confirm dialog show queued pricing and realistic timing ("~5-15 min (task queue)") when the toggle is active, live pricing otherwise. Saving config edits now also invalidates the cached cost estimate so the header price updates without a reload (a small pre-existing staleness bug that also affects the default setup).

`compose.yaml` forwards both variables into the self-host container; `docs/SELF_HOSTING_CLOUDFLARE.md` documents them.

## Testing

- New unit tests for `resolveRankCheckMethod` / `resolveQueueLiveFallback` (7 cases).
- `pnpm ci:check`, `pnpm test:ci`, `pnpm vite build` all pass.
- Verified end-to-end on a self-hosted Docker install: with the toggles set, a manual 140-keyword check ran fully through the queue (140 posted, 140 collected, 0 live fallbacks, ~6 min) at queued cost; without the variables, behavior is byte-identical to main.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -22,6 +22,9 @@ services:
       # Optional: AI features (SAM, onboarding chat). Without this the setup
       # gate in the app asks for the key, but nothing here would forward it.
       - OPENROUTER_API_KEY=${OPENROUTER_API_KEY:-}
+      # Optional rank-tracking cost controls: docs/SELF_HOSTING_CLOUDFLARE.md
+      - RANK_CHECK_MANUAL_METHOD=${RANK_CHECK_MANUAL_METHOD:-}
+      - RANK_CHECK_QUEUE_LIVE_FALLBACK=${RANK_CHECK_QUEUE_LIVE_FALLBACK:-}
       # Optional: Google Search Console. See
       # docs/SELF_HOSTING_GOOGLE_SEARCH_CONSOLE.md
       - GOOGLE_CLIENT_ID=${GOOGLE_CLIENT_ID:-}
--- a/docs/SELF_HOSTING_CLOUDFLARE.md
+++ b/docs/SELF_HOSTING_CLOUDFLARE.md
@@ -50,6 +50,11 @@ Copy the template and fill in the required values:
 cp .env.selfhost.example .env.selfhost
 ```
 
+Optional rank-tracking cost controls (also work as Worker variables):
+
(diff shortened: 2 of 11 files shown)


bensenescu:
Hey,

Thanks for contributing!

https://openseo.so/roadmap

_Contributing through Issues_

redrammedia:
Hi there,

Thanks for getting back to me. 
I have built so many features and add-ons to OpenSEO that I was actually thinking about creating pull requests, but I think it might be too much. 
So, I will keep them for myself for now and we'll just see how it turns out. Once I see if it works correctly, then I might get back to you.

Because I added many extra features, I found a couple of bugs. 
One of them I have already showed, but I think you fixed it already. 

We will see what's going to happen and, if everything turns out fine, I will set these new features up as pull requests later on.

Good luck and thanks for the hard work.

Best regards
Daniel

PR #182, opened 2026-07-31
fix(e2e): run keyword research specs without a DataForSEO API key

BittuBarnwal7479:
### Problem

The keyword research E2E tests use fixture data when `VITE_E2E_KEYWORD_FIXTURES` is enabled, so they shouldn't require DataForSEO credentials. However, two code paths still required `DATAFORSEO_API_KEY`, causing `pnpm test:e2e` to fail on a fresh checkout.

1. **`getSerpAnalysis` had no fixture support**

   * Unlike `researchKeywords`, it always called `KeywordResearchService`.
   * Without `DATAFORSEO_API_KEY`, the server failed with:

     ```
     Missing required environment variable: DATAFORSEO_API_KEY
     ```

2. **`getSeoApiKeyStatus` reported the API key as missing**

   * Even with fixtures enabled, it returned `configured: false`.
   * This showed the **"One quick setup step / Add your DataForSEO API key"** modal.
   * The modal's full-screen overlay blocked Playwright from clicking the page:

     ```
     <div ... class="fixed inset-0 z-50 ..."> subtree intercepts pointer events
     ```
   * All three tests timed out on tab clicks because the modal was covering the page.

### Fix

* Added `getSerpAnalysisFixture` alongside `getKeywordResearchFixture`.
* Updated `getSerpAnalysis` to use fixture data when E2E fixtures are enabled.
* Updated `getSeoApiKeyStatus` to return `configured: true` when either E2E fixture flag is enabled.

This follows the existing pattern used in `serverFunctions/keywords.ts` and `serverFunctions/domain.ts`.

Production behavior is unchanged because the fixture flags are only enabled by `playwright.config.ts`.

### Verification

Ran without `DATAFORSEO_API_KEY`:

```bash
pnpm exec playwright test e2e/keyword-research-navigation.spec.ts
```

Result:

```
3 passed (44.7s)
```

Also verified:

* ✅ `pnpm test:ci` (768 passed)
* ✅ `tsc --noEmit`
* ✅ `oxlint . --type-aware`

### Notes for Maintainers

While working on this, I noticed two unrelated issues that I can send as separate PRs:

1. **Windows E2E startup**

   * `playwright.config.ts` uses POSIX-style inline environment variables:

     ```bash
     NODE_OPTIONS= AUTH_MODE=... pnpm exec vite dev
     ```
   * This doesn't work in `cmd.exe`, so `pnpm test:e2e` cannot start on Windows.
   * Moving these variables into Playwright's `env` block makes it work cross-platform.

2. **Line endings on Windows**

   * The repository has no `.gitattributes` file.
   * On Windows with `core.autocrlf=true`, `prettier --check` reports changes for 811 files, making `pnpm ci:check` difficult to use locally.
   * Adding a `.gitattributes` file would keep line endings consistent across platforms.

### DIFF
--- a/e2e/fixtures/keyword-research-fixtures.ts
+++ b/e2e/fixtures/keyword-research-fixtures.ts
@@ -1,4 +1,4 @@
-import type { KeywordResearchRow } from "@/types/keywords";
+import type { KeywordResearchRow, SerpResultItem } from "@/types/keywords";
 import type { ResolvedResearchKeywordsInput } from "@/types/schemas/keywords";
 
 const MONTHLY_SEARCHES = [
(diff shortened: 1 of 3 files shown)


bensenescu:
Hey,

Thanks for contributing!

https://openseo.so/roadmap

_Contributing through Issues_

PR #184, opened 2026-07-31
fix(chat): stop yanking the transcript back down while the reader scrolls up

### LINKED ISSUE #160: Not able to scroll until SAM complete it's response

Nordalux:
`SamConversation` and `OnboardingChatConversation` both pinned their scroll container to the bottom on every `messages` change. During a streaming reply that fires per token, so scrolling up to re-read something is undone within milliseconds. Fixes #160, which reports it for SAM; the onboarding chat carried the same three lines.

The pin now holds only while the reader is still at the bottom — scrolling away releases it, scrolling back re-arms it. The duplicated effect moves into one hook, with the pin decision as a pure function so it is testable without a DOM.

Verified in:
- Windows 11, Node 24.15.0, pnpm 10.30.1 — `pnpm ci:check` clean, `pnpm test:ci` 774 passed, `pnpm vite build` green
- By hand in the running app — scrolling up mid-reply now stays put, and scrolling back to the bottom re-arms the pin.

### DIFF
--- a/src/client/features/onboarding/OnboardingChatConversation.tsx
+++ b/src/client/features/onboarding/OnboardingChatConversation.tsx
@@ -1,13 +1,14 @@
 import { useAgent } from "agents/react";
 import { useAgentChat } from "@cloudflare/ai-chat/react";
 import { useCustomer } from "autumn-js/react";
-import { useEffect, useRef, useState } from "react";
+import { useEffect, useState } from "react";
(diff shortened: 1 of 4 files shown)


bensenescu:
Hey,

Thanks for contributing!

https://openseo.so/roadmap

_Contributing through Issues_

PR #187, opened 2026-08-01
Library batch 2: plays 05-08, completing the Keyword Research library

jeremypenguin-max:
Batch 2 of the Keyword Research Strategy Library, opening with play 05. Kept separate from #185 (the blog post) so each can be reviewed on its own.

## Why this one first

The pillar already ships all eight cards, and four of them render with a "Next up" badge and no link. Play 05 is the one the pillar's own body copy already points at:

> The plays in this library replace it with three honest sources: your customers' language (play 01), Google's own question surfaces (play 02), and **your Search Console reality (play 05)**.

That sentence is live today and sends readers to a card they cannot click. This PR gives it a destination and removes the badge.

## What the page covers

- Striking-distance keywords (positions 11-30), grouped by ranking page rather than sorted by query, so the output is a work queue instead of a keyword list
- The 16-month retention window, and what a 90-day export hides
- Zero-click pages: heavy impressions, no clicks, and how to tell a title problem from a topic-drift problem
- A closing MCP prompt that runs all three

It cross-links `/blogs/dark-queries` as the why to this page's how, plus `/google-search-console-mcp`, and adds reciprocal links from the long-tail and clustering spokes.

## On the quotes

I verified these against the source recordings rather than my episode notes, and the check caught a misattribution I would otherwise have shipped. The "degapinator" framing and the triangulation idea are the host's, not the guest's. Grant Simmons is quoted only for what he actually said: the confidence-score line and the 16-month point. His skepticism is included too, since a page about closing a data gap is more honest with the practitioner who doubts it can be fully closed.

## Checks

- `npm run types:check` passes
- `npm run build` succeeds; the route prerenders and the sitemap picks it up
- `prettier --check` passes on all touched files, including the regenerated `routeTree.gen.ts`

## Not in this PR

No screenshot yet. Batch-1 spokes each carry a real app screenshot, and this one should get `gsc-striking-distance-openseo.png` (the ranked-keywords view filtered to positions 11-30) before merge. Happy to add it in a follow-up commit here. Plays 06, 07 and 08 are specced and follow next.


jeremypenguin-max:
Pushed plays **06** and **08** onto this branch, so the PR now covers three of the four outstanding library spokes.

| Play | Page | Targets |
|---|---|---|
| 05 | `gsc-programmatic-discovery` | striking-distance keywords, the 16-month window, zero-click pages |
| 06 | `opportunity-sizing-forecasting` | sizing a cluster, three-scenario forecasts, revenue per unit of difficulty |
| 08 | `positioning-to-demand` | validating a category name against real search demand |

Only play 07 (intent beyond Google) still renders as "Next up" on the pillar.

**Cross-links added both ways:** seeding to positioning, clustering to forecasting, forecasting to Search Console discovery, long-tail and clustering to play 05.

**Checks:** `types:check` passes, `build` succeeds with all three routes prerendering, sitemap now at 43 URLs, `prettier --check` clean including the regenerated `routeTree.gen.ts`.

**On play 08 and its keyword data.** I ran the research before writing, and the honest result is that this page has no strong cluster behind it. "brand positioning" (1,900/mo, KD 37) belongs to a marketing-strategy audience looking for positioning-statement templates, not to anyone validating search demand, and the surrounding low-difficulty terms are all template and example queries. Targeting them would pull the wrong reader and put the page in a fight with a large content ecosystem it has no business entering. So 08 is written as the library's closer: it earns its place through the argument and the internal links rather than through search volume. That is the same judgement the page itself recommends, which felt like the right way to resolve it, but say the word if you would rather it chase the volume.

**Still outstanding:** screenshots. Batch-1 spokes each carry a real app screenshot and these three do not yet.

jeremypenguin-max:
Play **07** is in, which completes the library. All eight cards on the pillar now carry an `href`, so the "Next up" badge no longer renders anywhere.

| Play | Page |
|---|---|
| 05 | `gsc-programmatic-discovery` |
| 06 | `opportunity-sizing-forecasting` |
| 07 | `intent-beyond-google` |
| 08 | `positioning-to-demand` |

Play 07 covers Pinterest as a search surface, the LinkedIn profile as the ranked asset, and the AI answer layer where rank and traffic have separated. Christopher Gimmer's point that an expanded paid block puts a number-one organic result fifth on the page is the one I would highlight, since it argues for CTR curves drawn from the real SERP rather than a textbook one.

`types:check` passes, `build` succeeds with all four routes prerendering, sitemap now at 44 URLs, `prettier --check` clean.

**One thing left: screenshots.** Batch-1 spokes each carry a real app screenshot and these four do not. Requested views, all destined for `web/public/library/keyword-research/`:

- `gsc-striking-distance-openseo.png` — ranked keywords filtered to positions 11-30
- `keyword-difficulty-volume-openseo.png` — research results sorted by difficulty against volume
- `intent-by-surface-openseo.png` — research results with the intent column visible
- `positioning-demand-validation-openseo.png` — a research query returning little or no volume

The last one is the awkward request: it needs a search that comes back near-empty, which is the page's whole argument. A real invented category term would do it.

Note the `play.href ? null : badge` branch in `index.tsx` is now unreachable. I left it in place rather than deleting it, on the assumption the library grows.

jeremypenguin-max:
All four screenshots are in, so this batch is complete. Every play now carries a real app view rather than a diagram, matching the batch-1 standard.

| Play | Image | What it shows |
|---|---|---|
| 05 | `gsc-striking-distance-openseo.png` | a keyword list filtered to positions 10-13 |
| 06 | `keyword-difficulty-volume-openseo.png` | results sorted by opportunity score, not volume |
| 07 | `intent-by-surface-openseo.png` | the intent column across a mixed result set |
| 08 | `positioning-demand-validation-openseo.png` | a category term with a nine-keyword universe |

Two of them turned out better than specified.

**Play 06.** The capture came back sorted by opportunity score, which is a stronger image than the difficulty-against-volume view I had planned. A term doing ten searches a month scores highest in the list; a term doing 6,600 sits several rows below it. That is the page's argument visible in the product instead of asserted in prose, so the caption now points at it directly.

**Play 08.** This one runs `fractional creative director`, which is the exact positioning George Little's earlier company failed on and the subject of the quote directly above it. The tool returns nine keywords in total. The term itself does 140 searches a month, everything around it is informational or navigational, and the only row with meaningful volume is a company name. Nothing in the list belongs to somebody trying to hire the role. The image and the interview now make the same argument with the same example, which is the best outcome this page could have had.

**One deliberate edit.** The play 05 capture originally included the URL column, which mapped each keyword to a live case-study URL on a client property. This is a public repository serving a public site, so I cropped the column before committing. The keywords and positions carry the argument on their own and the client's URL structure is not published.

PR #188, opened 2026-08-02
feat(audit): flag indexable pages with no structured data

ThomasMendy:
Site audits have no signal for structured data today, even though the crawler already records `hasStructuredData` on every page. A page that could be earning rich results — and that AI search engines otherwise have to interpret from prose alone — passes the audit silently.

This adds a `missing-structured-data` issue, reported for indexable HTML pages whose markup carries no JSON-LD, Microdata, or RDFa.

**Severity: `info`.** Missing markup is a missed opportunity rather than a defect, which matches how `noindex-page` and `deep-page` are treated.

**Noindex pages are skipped.** They cannot earn a rich result, so flagging them would only add noise to reports.

No new crawler work: the reporter reads a field `page-analyzer` already populates, so audits cost exactly what they did before.

### Testing

Four tests cover the new reporter: markup absent, markup present, noindex, and non-HTML. The shared `makePage` fixture now defaults to `hasStructuredData: true` so it keeps representing a healthy page.

- `pnpm test:ci` — 772 tests across 94 files, all passing
- `pnpm ci:check` — prettier, knip, both `tsc` configs, oxlint type-aware: clean
- `pnpm vite build` — clean

### DIFF
--- a/src/server/lib/audit/issues/page-reporters.test.ts
+++ b/src/server/lib/audit/issues/page-reporters.test.ts
@@ -45,7 +45,7 @@ function makePage(overrides: Partial<CrawledPageResult>): CrawledPageResult {
     imagesMissingAlt: 0,
     images: [],
     links: [HEALTHY_LINK],
-    hasStructuredData: false,
+    hasStructuredData: true,
     hreflangTags: [],
     isIndexable: true,
     responseTimeMs: 200,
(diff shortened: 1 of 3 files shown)


bensenescu:
Hey,

Thanks for contributing!

https://openseo.so/roadmap

_Contributing through Issues_

PR #189, opened 2026-08-03
fix: keep org slugs free of trailing dashes after truncation

Arunendra21:
**What**: Fixes a bug in `slugify` (`src/server/auth/org-slug.ts`).

`slugify` stripped leading and trailing dashes before truncating the value to 48 characters. When the 48 character cut lands on a `-` separator, the final slug still ends with a dash, which is exactly what the trailing-dash strip is meant to prevent.

For example, a long organization name can produce a slug like `my-really-long-org-name-that-goes-on-` (trailing dash), and in `default-hosted-organization.ts` that becomes `...goes-on--<suffix>` with a double dash once the suffix is appended.

**Fix**: truncate to 48 characters first, then strip leading and trailing dashes, so the slug never keeps a stray dash on either end. Normal inputs are unchanged.

**Testing**: added `src/server/auth/org-slug.test.ts` with vitest cases for normal values, edge trimming, the empty fallback, the truncation-on-separator case, and `toHex`.

### DIFF
--- a/src/server/auth/org-slug.test.ts
+++ b/src/server/auth/org-slug.test.ts
@@ -1,0 +1,34 @@
+import { describe, expect, it } from "vitest";
+
+import { slugify, toHex } from "@/server/auth/org-slug";
+
+describe("slugify", () => {
+  it("slugifies normal values", () => {
+    expect(slugify("My Workspace")).toBe("my-workspace");
+    expect(slugify("  Hello, World!  ")).toBe("hello-world");
+  });
+
+  it("strips leading and trailing separators", () => {
+    expect(slugify("---weird---")).toBe("weird");
+  });
+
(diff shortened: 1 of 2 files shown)


bensenescu:
Hey,

Thanks for contributing!

https://openseo.so/roadmap

_Contributing through Issues_

Arunendra21:
Thanks for the heads-up, and no worries at all — completely understand wanting to keep the roadmap moving. Happy to refile this as an Issue following the contributing guide if that is the more useful path right now. Appreciate the work on open-seo!

Arunendra21:
Closing this per the updated contributing guidelines — makes sense to keep the roadmap moving. I'll refile it as an Issue so it can be prioritized properly. Thanks again @bensenescu!

PR #193, opened 2026-08-06
Search Intelligence phase 2 — keyword gap, rank tracking, share of search

mattia-chksg:
Engine side of Phase 2. Eight new tables in both dialects (migration `0038` /
pg `0015`, all additive), the keyword/cluster stores, gap classification,
Tracked Keyword Share of Search, ranking-event detection and a Phase-2 cost
ledger with its own cost centre.

Still fixture-only: `runRankTick` refuses a `live` provider explicitly, since
this engine has no DataForSEO credential and must never reach for Morgana's
Brand Monitoring one.

### Notable

- **`si_rank_snapshots`, not `rank_snapshots`.** Upstream already owns that
  name; sharing it made drizzle silently drop our table from the migration
  while our export won the runtime barrel, so writes would have landed in
  upstream's table.
- **Losses require confirmation, gains do not.** A drop is only recorded as an
  event when the observation before the previous one was healthy — a single bad
  observation is usually SERP noise. Competitive events fire for `critical`
  keywords only.
- **Share of search refuses below 50% volume coverage** and returns
  `insufficient_data` rather than a misleading number.
- **`ranking-events/ack`** exists because Morgana owns delivery; without an
  acknowledgement round-trip the same event is handed out every tick.
- Dedupe keys are UNIQUE constraints, so replaying a tick is free.

### Verification

typecheck, oxlint `--type-aware` and knip clean; 126 tests in
`src/server/morgana`; 180 schema-parity assertions; migration applied to the
staging D1 (25 commands, 0 destructive); deployed and verified over a real
Service Binding with no public ingress and zero spend.


mattia-chksg:
Opened against the wrong repository by mistake — this belongs on our fork. Closing immediately.

PR #196, opened 2026-08-07
feat(mcp): add rank-tracking keyword write tools

Despiram:
## Motivation

Self-hosted agents can already read the rank tracker over MCP (`get_rank_tracker`), but they cannot manage its watchlist — adding or pruning tracked keywords still requires the dashboard. This enables agent-driven keyword management: an agent doing keyword research over MCP can now promote its findings straight into the tracker (and clean up stale ones) without a human round-trip.

## What this adds

- **`add_rank_tracking_keywords`** — adds keywords (1–100) to a project's rank tracker. Reports added, already-tracked (duplicate), and cap-skipped keywords. Uses no credits and triggers no check; positions appear on the next scheduled or dashboard-triggered check.
- **`remove_rank_tracking_keywords`** — removes keywords from the watchlist, matching case-insensitively after trimming (the stored normal form). Unmatched keywords are reported rather than erroring. Marked `destructiveHint: true`.
- Both tools accept an optional `trackerId` and resolve the tracker automatically when the project has exactly one; with several trackers, the error lists them so the agent can pick.
- Both are thin wrappers over the existing `RankTrackingService.addKeywords` / `removeKeywords` — no duplicated business logic — and are scoped through `withMcpProjectAuth` like the other project tools.
- Registered in `src/server/mcp/server.ts` and listed on the AI/MCP settings page (`AvailableTools`).

Intentionally left out: auto-triggering a rank check after adding (the server-function path does this for the dashboard). An MCP write should not silently spend credits; the tool says so in its output text instead.

## How it was tested

- New `src/server/mcp/tools/rank-tracking-write-tools.test.ts` (7 tests) following the existing tool-test pattern: service pass-through + duplicate reporting, single-tracker resolution, multi-tracker/unknown-tracker rejection before any write, cap-skip reporting, keyword→ID mapping on remove, and no service call when nothing matches.
- Full suite: `pnpm test` — 95 files, 775 tests, all passing.
- `knip`, `tsc --noEmit` (main + badseo), and `oxlint . --type-aware` all clean; new/touched files formatted with Prettier. (Repo-wide `prettier --check` fails on this Windows checkout due to CRLF conversion, unrelated to this change.)

### DIFF
--- a/src/client/features/ai-mcp/AvailableTools.tsx
+++ b/src/client/features/ai-mcp/AvailableTools.tsx
@@ -24,6 +24,16 @@ const toolCategories: ToolCategory[] = [
         description: "Read tracked keyword positions.",
       },
       {
+        name: "add_rank_tracking_keywords",
+        title: "Add rank tracking keywords",
+        description: "Add keywords to a rank tracker's watchlist.",
+      },
+      {
+        name: "remove_rank_tracking_keywords",
+        title: "Remove rank tracking keywords",
+        description: "Remove keywords from a rank tracker's watchlist.",
+      },
+      {
         name: "get_keyword_metrics",
         title: "Get keyword metrics",
         description:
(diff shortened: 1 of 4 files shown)


bensenescu:
Hey,

Thanks for contributing!

https://openseo.so/roadmap

_Contributing through Issues_

PR #203, opened 2026-08-09
fix(landing-page): make GitHub star count live

carterwsmith:
## Summary

Makes the GitHub star count in the marketing-site navigation dynamic, currently it isn't.

The existing implementation fetched the count during prerendering, but included it with each deployment (at time of posting it is stuck at `9.9k`) and fell back to a hard-coded `2.1k`. This change refreshes the count in the browser so it can update without a new deployment.

## Changes

- Fetch the latest star count from the GitHub API after hydration.
- Remove the hard-coded `2.1k` fallback, instead hiding the count if GitHub is unavailable.
- Use the existing prerendered count on the initial page render.

## Verification

- `pnpm exec prettier --check src/routes/_marketing.tsx`
- `pnpm run types:check`
- `pnpm run build:app`


bensenescu:
Do you know what the github rate limits are? Concerned about getting throttled. 

I wonder if we could do this server side in Tanstack or cache it somehow to that its more up to date.

I've had it on my todo list to just auto deploy every day to fix this.

carterwsmith:
I didn't consider how low it could be, just 60/hr! I'll see if there's an elegant solution that doesn't require redeploys

PR #204, opened 2026-08-09
feat(audit): AEO checks — schemamap and JSON-LD linked graph (part 2 of #99)

Santofer:
Part 2 of #99 — the two checks #122 explicitly leaves open.

@WAHIB-EL-KHADIRI offered these two in [his comment](https://github.com/every-app/open-seo/issues/99#issuecomment-5043694362) after shipping the other three in #122, so this PR takes them and touches nothing #122 owns conceptually. There is file overlap (`audit-issues.ts`, `page-analyzer.ts`, `types.ts`, `siteAuditWorkflowPhases.ts`) — whoever lands second rebases, and I'm happy for that to be me.

### `jsonld-not-linked-graph` — page-level, `info`

Fires when a page publishes **two or more independent JSON-LD entities that never reference each other**.

Each block can be perfectly valid on its own; that's the point. Consumers merge structured data into one graph and then ask it questions — who wrote this, who published it — and three disconnected islands answer none of them. The fix is a single `@graph`, or a stable `@id` per entity referenced from the others.

Never fires on: one entity, a `@graph`, a resolvable cross-block `@id`, malformed JSON, or a non-HTML response.

### `schemamap-missing` — site-level, `info`

Fires when the site serves no `/schemamap.xml` **and** names none via a `Schemamap:` directive in robots.txt. Absence is an opportunity, not a defect (the convention is young), hence `info`.

`robots-parser` only surfaces directives it knows about, so the directive is read off the raw robots.txt text the discovery phase already checkpoints — no second fetch of robots.txt.

## Two structural choices worth reviewing

**JSON-LD is folded into counters, never buffered.** `page-analyzer.ts` exists because a full DOM per page was the engine's dominant OOM cause at 25 concurrent parses on a 128MB isolate, so I did not want to reintroduce that by accumulating raw JSON-LD strings. Blocks are folded one at a time into five scalars (`blocks`, `nodes`, `hasGraph`, `crossReferenced`, `truncated`) — at most one block's text is live at any moment. Blocks past the size/count caps mark the summary `truncated`, which **suppresses** the check rather than guessing from partial data.

The `jsonLd` field is transient like `isHtml`/`htmlBytes`: `runPageReporters` runs on the in-memory result inside the crawl chunk, so **no DB migration and no new persisted column**.

The cheerio parity reference in `page-analyzer.test.ts` folds through the same pure accumulator, so `expect(streamed).toEqual(reference)` still compares tokenizer against DOM and nothing else.

**`schemamap-missing` is the engine's first site-level issue.** Everything before it hangs off a page row, so this needed a small amount of new shape:

- It emits `pageId: null` with the origin as `pageUrl`. The issues table already allows a null page reference (`pageId` is not `.notNull()`), and `pageUrl` is required, so the origin fills it.
- It rides **inside the existing `multipage-checks` step** rather than adding a phase: same checkpoint, and `insertIssues` is idempotent via `deterministicAuditRowId` + `onConflictDoNothing`, so a retry that re-fetches `/schemamap.xml` cannot duplicate rows.
- `finalizeAudit` now receives `origin` and the checkpointed `robotsText`.

The impure entry point (`runSiteChecks`) is one thin function; every decision lives in a pure predicate, mirroring how `discovery.ts` splits `fetchRobotsTxtText` from `parseRobotsTxt`.

One judgement call I'd flag: a network error fetching `/schemamap.xml` is treated as "not served", so a site whose root is briefly unreachable can surface the issue. The alternative — suppress on error — hides it permanently for anyone with a flaky root. For an `info` finding I think a rare false positive beats a silent false negative, but say the word and I'll flip it.

## badseo

New **Structured data** category with a matched pair:

- `/structured-data/unlinked-graph` — Article + Organization + Person as three islands → `["jsonld-not-linked-graph"]`
- `/structured-data/linked-graph` — the same three entities joined by `@id` → `[]`

The second one is the point: it's the regression guard that stops this check firing on correct markup, and it's realistic (separate blocks emitted by separate components, agreeing on identifiers).

Site-level issues have no fixture to hang off, so the harness asserts `schemamap-missing` directly (modeled on the existing trailing-slash guard) and lists it in a `SITE_LEVEL_ISSUES` coverage exemption — asserted, so the exemption isn't a lie. `crawl()` now returns `robotsText` so the site check can reuse it.

## Verification

| Gate | Result |
| --- | --- |
| `tsc --noEmit` | clean |
| `tsc --noEmit -p badseo/tsconfig.json` | clean |
| `oxlint --type-aware` | 0 warnings, 0 errors |
| `vitest run` | **1013 passed** (124 files) |
| badseo harness, live instance | **44/44 checks**, issue-type coverage **29/29** |

Harness output for the new rows:

```
✓ Structured data is not a linked graph           jsonld-not-linked-graph
✓ Structured data as a linked graph               clean
✓ Site: no schemamap declared                     schemamap-missing reported
```

`knip` still fails on `vite.config.ts` (`Cannot read properties of undefined (reading 'join')`) — pre-existing on a clean `main`, unrelated to this change.

### DIFF
--- a/badseo/scripts/run-audit.ts
+++ b/badseo/scripts/run-audit.ts
@@ -17,6 +17,7 @@ import {
   discoverUrls,
   parseRobotsTxt,
 } from "../../src/server/lib/audit/discovery";
+import { runSiteChecks } from "../../src/server/lib/audit/issues/site-checks";
 import {
   normalizeUrl,
   isSameOrigin,
@@ -68,6 +69,7 @@ async function crawl(origin: string): Promise<{
   pages: CrawledPageResult[];
   links: CrawlLink[];
   completed: boolean;
+  robotsText: string | null;
 }> {
   const { urls: sitemapUrls, robotsText } = await discoverUrls(
     origin,
@@ -146,7 +148,7 @@ async function crawl(origin: string): Promise<{
   }
 
   const completed = linkQueue.length === 0 && sitemapQueue.length === 0;
-  return { pages, links, completed };
+  return { pages, links, completed, robotsText };
 }
 
 /** In-memory equivalents of the two D1-backed multipage checks. */
@@ -243,7 +245,7 @@ async function main() {
   await warmup();
   const origin = new URL(BASE).origin;
   const startUrl = normalizeUrl(`${origin}/`) ?? `${origin}/`;
-  const { pages, links, completed } = await crawl(origin);
+  const { pages, links, completed, robotsText } = await crawl(origin);
 
   if (process.env.DEBUG_DEPTH) {
     for (const p of pages) {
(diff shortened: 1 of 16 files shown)


bensenescu:
Hey,

Thanks for contributing!

https://openseo.so/roadmap

_Contributing through Issues_

PR #213, opened 2026-08-11
test(docker): guard entrypoint build fingerprint against vite envPrefix drift

WAHIB-EL-KHADIRI:
## What

Adds a fast unit test (`src/shared/docker-entrypoint-env-sync.test.ts`) asserting that the env allow-list used by `docker-entrypoint.sh` to fingerprint the client build stays in sync with `vite.config.ts`'s `envPrefix` — plus the build-only extras (`POSTHOG_SOURCEMAPS`).

## Why

`docker-entrypoint.sh` skips the client build when the build-relevant envs are unchanged:

```sh
FINGERPRINT="$(env | grep -E '^(VITE_|AUTH_MODE|BYPASS_EMAIL_VERIFICATION|POSTHOG_PUBLIC_KEY|POSTHOG_HOST|TURNSTILE_SITE_KEY|POSTHOG_SOURCEMAPS)' | sort | sha256sum ...)"
```

That list must mirror `vite.config.ts` `envPrefix` (the prefixes Vite inlines into the client bundle). Today the coupling is enforced only by a "keep in sync" comment. If a new `envPrefix` entry is added without updating the entrypoint, a self-host container could **silently reuse a stale build** after a build-relevant env changed. This test makes that drift a loud CI failure instead.

## Notes

- Pure config-integrity test: reads both files, compares the two sets. No runtime/build deps.
- Verified locally: `pnpm exec vitest run src/shared/docker-entrypoint-env-sync.test.ts` → 1 passed.

### DIFF
--- a/src/shared/docker-entrypoint-env-sync.test.ts
+++ b/src/shared/docker-entrypoint-env-sync.test.ts
@@ -1,0 +1,37 @@
+import { readFileSync } from "node:fs";
+import { describe, expect, it } from "vitest";
+
+/**
+ * `docker-entrypoint.sh` skips the client build when the build-relevant envs are
+ * unchanged, fingerprinting them with a `grep` allow-list. That list must stay in
+ * sync with `vite.config.ts`'s `envPrefix` (the prefixes Vite inlines into the
+ * client bundle) plus the build-only extras that move the output (POSTHOG_SOURCEMAPS
+ * switches the outDir). If they drift, a container could silently reuse a stale
+ * build after a build-relevant env changed — this test fails loudly instead of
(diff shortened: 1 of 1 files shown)


bensenescu:
Hey,

Thanks for contributing!

https://openseo.so/roadmap

_Contributing through Issues_

WAHIB-EL-KHADIRI:
Understood, and no argument from me on the reasoning in the contributing guide.

Filed the substance as #316 so it's in the format you asked for. Closing this, but leaving the branch up - the test is one file if you or your agent want it as a starting point.

PR #214, opened 2026-08-12
feat(mcp): let AgentOnboard agents authenticate into the MCP server

atpaawej:
## What this gives you

This PR lets any **AgentOnboard**-connected AI agent call OpenSEO's existing MCP server directly — no MCP client registration, no OAuth consent dance, no API key. The agent sends a short-lived session token and acts as the matching OpenSEO user.

**The agent experience (no setup for the human):**
1. A user tells their agent to "check our keyword rankings on OpenSEO."
2. The agent mints a 5-minute session token with `aon token get` (AgentOnboard CLI).
3. The agent calls `POST /mcp` with `x-session-token: <token>` and runs the standard MCP `tools/list` + `tools/call` handshake.
4. OpenSEO verifies the token once against AgentOnboard's `/api/verify`, maps the AgentOnboard email to the user's OpenSEO account, and serves the request with **exactly the same tools and permissions as that user's own OAuth'd MCP client**.

**What's added:**
- **A new auth path into the existing `/mcp` server.** An additive, header-based entry (`x-session-token`) that synthesizes the same MCP auth context the self-hosted path already uses. It's opt-in and env-gated — set `AGENTONBOARD_PARTNER_KEY` (hosted deployments) and it's live; unset, the header is ignored and the existing OAuth flow is untouched.
- **No new REST API, no changes to server functions, dashboard routes, or existing auth.** This is strictly an additional way into the MCP surface OpenSEO already exposes to agents.
- **Email is the join key.** The returned AgentOnboard email maps to the matching OpenSEO account (unique on email). No matching account → a clear rejection. Unverified email → rejected. The agent runs in the user's own organization.
- **Fail-closed and classified.** Bad/expired token → 401. Revoked partner key or AgentOnboard outage → server-side error (logged), never blamed on the agent. Requests without the header fall through to normal auth.
- **Tests + verified contract.** 23 new tests (service + handler); full suite green. The `/api/verify` contract was exercised against the live AgentOnboard API.

**What it does NOT do (by design):**
- No new endpoints, no agent-visible REST API, no tool-scoping — an AgentOnboard agent is equivalent to the user's own client.
- Agent-call logging is a known follow-up, not in this PR.

### Why this is worth integrating
It's the AgentOnboard distribution channel: agents discover OpenSEO through the AgentOnboard partner directory, and can act on OpenSEO with zero per-agent setup. One integration, any agent.

---

**Testing:** `tsc --noEmit`, `oxlint --type-aware`, `vitest run` (full suite), `knip`, `prettier --check`, and `vite build` all pass. Live `/api/verify` round-trip confirmed against the real AgentOnboard API.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -18,6 +18,7 @@ import {
 import { requestWithPublicOrigin } from "@/server/mcp/public-origin";
 import { MCP_ROUTE } from "@/server/mcp/context";
 import { handleSelfHostedOpenSeoMcpRequest } from "@/server/mcp/transport";
+import { handleAgentOnboardMcpRequest } from "@/server/agentonboard/agentonboard-mcp";
 import { withPgClient } from "@/db";
 import {
   AUTUMN_WEBHOOK_PATH,
@@ -136,11 +137,11 @@ function fetch(
   return withPgClient(() => Promise.resolve(handleFetch(request, env, ctx)));
 }
 
-function handleFetch(
+async function handleFetch(
   request: Request,
   env: Env,
   ctx: ExecutionContext,
-): Response | Promise<Response> {
+): Promise<Response> {
   ctx.waitUntil(maybeSendSelfHostHeartbeat());
 
   const authMode = getAuthMode(env.AUTH_MODE);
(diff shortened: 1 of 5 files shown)


bensenescu:
We're not introducing new third party dependencies right now.

atpaawej:
Hi Ben,

Thank you for taking the time to look into this.

For a quick demonstration, please take a look at the video below to see how
AgentOnboard works. As you can see, I haven't connected any MCP or anything
named RankGeo, but the agent is still able to use RankGeo on my behalf by
directly authenticating into it. That is exactly what agentonboard.xyz does.

Everything is completely free while we are in early beta, and we do not
proxy any of your data. This is simply about providing convenience for your
users.

Best regards,
Aawej Pathan

atpaawej:
https://drive.google.com/file/d/14tce-nEtGpLPn95B2BJlL1plGKdYbWJd/view?usp=sharing

PR #236, opened 2026-08-19
fix(analytics): keep MCP API keys out of PostHog session replay

### LINKED ISSUE #233: New API key can be captured by session replay
## TL;DR

When a user creates an MCP API key, the full key is shown as plain text and can be captured by PostHog session replay. The key should be excluded from recordings.

## What is happening?

The one-time key reveal renders the key as text in `src/client/features/settings/ApiKeySettings.tsx`. Session recording is configured in `src/client/lib/posthog.ts` with `maskAllInputs: true` and `maskTextSelector: "[data-ph-mask], .ph-mask"`. The reveal element is not an input and carries neither selector, so where PostHog is enabled the recording can contain a working key. Keys do not expire by default, so one stays valid until it is revoked.

## What should happen?

The key reveal should be excluded from session recordings, including any captured network response body that contains the key.

## Extra context

This applies wherever `POSTHOG_PUBLIC_KEY` / `POSTHOG_HOST` are set, so the hosted app and any self-hoster who enables analytics. Self-hosted deployments without PostHog configured are not affected.

The rest of the API key handling looks sound: 64-character keys, hashed at rest, hashed lookup on verify, and plaintext returned only at creation. The gap is only that the one moment the plaintext is on screen is also recorded.

Unrelated but worth flagging: the repo currently has GitHub private vulnerability reporting disabled and no `SECURITY.md`, so there is no non-public way to send a report like this one. Enabling it would help.

0xenzyme:
## Summary
- Mask the one-time plaintext MCP API key reveal in Settings (`data-ph-mask` / `ph-mask`) so PostHog session replay cannot capture it from the DOM.
- Sanitize captured network requests: drop request/response bodies for Better Auth `/api/auth/api-key*` routes, and redact full `oseo_…` keys elsewhere.
- Fixes #233.

## Test plan
- [x] `pnpm test -- src/client/lib/posthog-sanitize.test.ts`
- [ ] In a hosted (or PostHog-enabled) build, create an API key in Settings and confirm the reveal still works for the user
- [ ] Confirm a session recording of that create flow does not include the plaintext key in the DOM or network payload

### DIFF
--- a/src/client/features/settings/ApiKeySettings.tsx
+++ b/src/client/features/settings/ApiKeySettings.tsx
@@ -197,7 +197,11 @@ export function ApiKeySettings() {
                   to <span className="font-mono text-xs">{mcpUrl}</span>.
                 </p>
                 <div className="mt-4 flex items-center gap-2">
-                  <code className="min-w-0 flex-1 overflow-x-auto rounded bg-base-200 px-2.5 py-2 font-mono text-xs">
+                  {/* Plaintext key is shown once; keep it out of PostHog replay. */}
+                  <code
+                    className="ph-mask min-w-0 flex-1 overflow-x-auto rounded bg-base-200 px-2.5 py-2 font-mono text-xs"
+                    data-ph-mask
+                  >
                     {createdKey}
                   </code>
                   <CopyButton
--- a/src/client/lib/posthog-sanitize.test.ts
+++ b/src/client/lib/posthog-sanitize.test.ts
@@ -1,11 +1,15 @@
 import { describe, expect, it } from "vitest";
 import {
+  sanitizePostHogCapturedNetworkRequest,
   sanitizePostHogProperties,
   sanitizePostHogUrl,
 } from "@/client/lib/posthog-sanitize";
 
 const consentUrl =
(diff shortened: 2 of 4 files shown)


Abhirup0 (commented):
Hey @0xenzyme, thanks for putting this together! 

The approach looks really solid to me. Adding both `data-ph-mask` and `.ph-mask` to the code block is great, and dropping the request/response bodies entirely for the Better Auth API key endpoints is much safer than just relying on regex pattern matching for the creation payload. Also like that the `{16,}` regex length check keeps the shortened `oseo_abcd` display prefixes intact in the UI table.

I just had a couple of small suggestions/thoughts for extra defense-in-depth if you think they're worth adding:

1. **Scrubbing keys from URL query params:**
   In `sanitizePostHogUrl`, we currently handle `/oauth-consent` and strip `email`. If someone ever passes an API key as a query param (like `?apiKey=oseo_...`), `sanitizePostHogCapturedNetworkRequest` might still leave it in `request.name`. Running the regex over `url.search` or `url.toString()` as well could be a nice safeguard.

2. **Relative URLs in `sanitizePostHogUrl`:**
   In `isApiKeyAuthRequestUrl`, you used `new URL(name, "https://example.invalid")` which handles relative paths cleanly. In `sanitizePostHogUrl`, `new URL(value)` without a base will throw on relative URLs (like `/api/projects`) and jump straight to the `catch` block returning the raw string. Passing a dummy base URL there too might make it more consistent.

3. **Request headers (minor):**
   If PostHog happens to record `requestHeaders` in your setup, it might be worth making sure `authorization` or `x-api-key` headers get scrubbed if they contain `oseo_...` tokens.

Either way, the core fix is super clean and the test cases cover the main paths nicely. Thanks again for the PR!

PR #315, opened 2026-09-09
fix(google-ads): offer reconnect from the empty account picker

tonyswartz:
## What

When Google Ads account discovery returns no candidates, the picker now shows a **Reconnect with Google** button, with copy pointing at the most common cause (the Google Ads permission checkbox left unchecked on Google's consent screen).

## Why

The empty state was a dead end: the Disconnect button only renders after a successful connect, and the card's Connect button is hidden once a grant exists. A user who skipped the consent checkbox (which surfaces server-side as `accountsUnavailable`, a 403 scope error, not the 401 that triggers `requiresReconnect`) had no way to redo the OAuth flow. Hit in production by Tony ~23:47Z tonight, minutes after first connect.

Reconnecting is safe and sufficient: the self-hosted OAuth callback upserts the grant by (user, provider, Google account id) and `prompt=select_account consent` forces the chooser + consent screen on every pass.

## Verification

- `prettier --check` clean, `tsc --noEmit` clean
- Full vitest suite: 1177 passed, 140 files


tonyswartz:
Opened against the wrong repo (meant for a downstream fork). Apologies for the noise.

PR #319, opened 2026-09-10
feat(google-ads): surface LSA credit_state, call duration, and already-submitted re-files

tonyswartz:
## Why

Tony asked for the remaining LSA improvements after 338 filed `FAIL_NOT_ELIGIBLE`: matcher training on both survey halves, credit-eligible reasons only when true, `credit_state` on the leads tool + weekly digest, and louder charged+unverifiable (conversation duration if Google will give it). Live 338 also showed `lead_feedback_submitted` stays false after a successful file; Google's real once-per-lead stop is `RESOURCE_ALREADY_EXISTS`.

## What

`get_local_services_leads` now returns:

- `creditState` (`local_services_lead.credit_details.credit_state` — the leaf is selectable on live v25; the parent stays off the SELECT per #9)
- `leadFeedbackSubmitted`
- `conversationDurationMillis` (longest phone-call duration; does **not** fetch `call_recording_url`)

Extra SELECT fields and the conversation query fall back to null on a 400 so the lead list still returns. The performance/dashboard path is unchanged.

`provide_lead_feedback` maps a re-file `RESOURCE_ALREADY_EXISTS` 400 to `lead_feedback_already_submitted` instead of "rejected report request."

Cron prompts (outside this repo, already updated on disk):

- File `VERY_SATISFIED` / `SERVICE_RELATED` (or `BOOKED_CUSTOMER`) on high-confidence charged matches
- Duplicate detection: same phone within 15 days → `DUPLICATE`
- Do not recode civil as spam
- Job-type/geo still file for matcher training; expect `FAIL_NOT_ELIGIBLE`
- Digest reads `creditState` (Google's ~30-day auto-review)
- Charged+unverifiable posts include duration

## Tests

At `3e6661ff68425745f65583983ed4d0f4d2c30128`: `tsc --noEmit` clean, `oxlint . --type-aware` clean, `vitest run` 143 files / 1216 tests.


tonyswartz:
Opened against the wrong remote — this is a self-host fork change for tonyswartz/open-seo, not an upstream contribution. Closing.

PR #350, opened 2026-09-20
Show the updates[] wrapper in project-context write examples

VodouAI:
Nine skills document the research-log write as a bare patch op:

```
append a research log entry: `{ appendResearchLog: { summary: "..." } }`
```

Read as the tool's argument shape — which is how it reads without the surrounding call — that produces a silent data loss. `update_project_context` takes `{ projectId, updates }`, and unknown top-level keys are stripped before the handler sees them. So an agent that sends `appendResearchLog` or `addKeyPages` at the top level alongside a valid `updates[]` gets **`Updated project context (1 change(s))`** while that write is discarded.

Verified against a running instance:

- top-level only, no `updates[]` → `Input validation error: updates: Invalid input: expected array, received undefined`. Loud, fine.
- valid `updates[]` plus a top-level `addKeyPages` → success, key page silently dropped.

The examples now show `{ updates: [{ appendResearchLog: { summary: "..." } }] }`.

`seo-project-setup` is deliberately untouched: its op list already sits under `update_project_context(projectId, updates)`, so the wrapper is implied there.

Found when an agent running `seo-audit` reported its research-log and key-page writes as successful and neither appeared.

Docs-only change — no behavior change. `sync-plugin-skills` re-run, idempotent, so the `plugins/openseo/skills` check stays green.


VodouAI:
The root cause here isn't really the skill wording — it's that unknown top-level keys are stripped silently. Every tool registers through `registerOpenSeoTool`, which normalizes inputs via `objectSchema`, so making input objects strict is a few lines in one place:

```ts
const normalized = objectSchema(tool.config.inputSchema);
inputSchema: normalized instanceof z.ZodObject ? normalized.strict() : normalized,
```

(Outputs would stay permissive, per the existing comment there.) That would turn this class of silent write-loss into a validation error across every tool, not just the ones documented here.

I didn't do it in this PR because it changes behavior for every client — anything currently sending a stray field starts getting a hard error instead of a quiet success, and you're better placed than I am to know whether that's true of any client in the wild. Happy to open it as a separate PR if you want it, either instead of or alongside this one.

